PR #14, opened 2015-03-28
Add erase operation

hamhands:
### What changed
* Added erase operation
* Removed spaces in name key value in package.json to fix error

### Screencap
new


svmatthews:
Thanks @hamhands! Confirmed it works on my end.

hamhands:
@svmatthews :raised_hands:

Issue #9, opened 2015-03-24
set up an http server Grunt task that works with 'watch'


svmatthews:
would be nice to get the default `grunt` command to roll a server and watch for changes in `src/`. What's the best way to do this?

alukach:
Completed in #18

PR #17, opened 2015-03-30
Remove python dep

knownasilya:
Make running this super easy, no python to worry about.

### DIFF
--- a/.bowerrc
+++ b/.bowerrc
@@ -1,0 +1,3 @@
+{
+  "directory" : "app/static/bower_components"
+}
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
 .DS_Store
 node_modules/
-bower_components/
+app/static/bower_components
 **/.sass_cache/**
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ bower install
 And run a simple server with
 
 ```
-python -m SimpleHTTPServer
+npm start
 ```
 
 You should now be able to access the application at `http://localhost:8000/app`
--- a/README.md
+++ b/README.md
@@ -68,4 +68,4 @@ And run a simple server with
 npm start
 ```
 
-You should now be able to access the application at `http://localhost:8000/app`
+You should now be able to access the application at `http://localhost:8000`
--- a/app/index.html
+++ b/app/index.html
@@ -2,13 +2,13 @@
 <head>
   <title>Drip. Drop.</title>
   <!-- bower:js -->
-  <script src="../bower_components/jquery/dist/jquery.js"></script>
-  <script src="../bower_components/turf/turf.min.js"></script>
-  <script src="../bower_components/mapbox.js/mapbox.js"></script>
+  <script src="static/bower_components/jquery/dist/jquery.js"></script>
+  <script src="static/bower_components/turf/turf.min.js"></script>
+  <script src="static/bower_components/mapbox.js/mapbox.js"></script>
   <!-- endbower -->
   <link href='http://fonts.googleapis.com/css?family=Open+Sans:400,700' rel='stylesheet' type='text/css'>
   <link rel="stylesheet" href="//maxcdn.bootstrapcdn.com/font-awesome/4.3.0/css/font-awesome.min.css">
-  <link rel="stylesheet" href="../bower_components/mapbox.js/mapbox.css">
+  <link rel="stylesheet" href="static/bower_components/mapbox.js/mapbox.css">
   <link href="static/css/app.css" rel="stylesheet">
 </head>
 <body>
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
   },
   "main": "src/main.js",
   "scripts": {
-    "start": "goat -e ./app/index.html ./ -p 8000"
+    "start": "goat -e app/index.html -p 8000"
   },
   "devDependencies": {
     "browserify": "2.35.4",
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
   },
   "main": "src/main.js",
   "scripts": {
-    "start": "goat -e app/index.html -p 8000"
+    "start": "goat -e ./app/index.html ./ -p 8000"
   },
   "devDependencies": {
     "browserify": "2.35.4",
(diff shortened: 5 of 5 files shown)


svmatthews:
Hey! Thanks for setting this up. Two things:

1. Right now if the server exists at `.../app` level there is no access to the root level `/bower_components/` directory. I like the idea that the user doesn't need to run it at `/app` though, so maybe we should set up a [`.bowerrc`](http://bower.io/docs/config/) file. I wonder how that works with `wiredep`?
2. Any interest in figuring out how to properly set this up to run with grunt `watch`? Relates to #9

knownasilya:
> Right now if the server exists at `.../app`

Can you explain what you mean by this statement?

svmatthews:
Yes of course. Since the server is being set within the `/app` directory via [this parameter](https://github.com/knownasilya/drop-n-chop/commit/f5d6d92985e0246330d98bede5d0e44517674ecd#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R13), the paths to the necessary vendor libraries (i.e. jQuery, turf, mapbox.js) are broken and do not load. This is a result of the `/bower_components` library being one level above the server and therefore inaccessible.

Here's an image of the console errors I'm seeing:
screen shot 2015-03-30 at 10 25 03 am

I think the best fix here would be to move the `/bower_components` into the `/app` directory (i.e. `/app/bower_components` or `/app/static/lib`) using the `.bowerrc` option.

knownasilya:
Seems to be working fine, on my end. I did `npm start` in the main directory and opened `localhost:8000`.

screen shot 2015-03-30 at 1 33 30 pm

svmatthews:
Hmm, I honestly don't know what's going on! I've run `npm install` and `bower install` and `npm start` and am still getting the errors. Any ideas?

knownasilya:
Try doing `rm -rf node_modules && npm cache clean && npm install`

svmatthews:
Didn't work for me :(

svmatthews:
I'm wondering if it would be best to use the `.bowerrc` file and do this through `grunt serve` instead, just to keep everything under the same task runner?

alukach:
I think the problem is that `goat` is only serving `app`.  This fix did it for me: https://github.com/knownasilya/drop-n-chop/pull/1

Side note: I did a PR to the PR, which is a bit weird. Felt that seemed a bit more sensible than creating a new PR. :confused:

knownasilya:
@alukach it still doesn't explain why it works on my end :smile:

I think PRs on PRs is a good thing :+1:

knownasilya:
@svmatthews give that a try.

alukach:
@knownasilya Yeah, I'm unsure how it could have worked for you. Are you sure that your browser wasn't serving cached data from `/bower_components`?  I was able to reproduce @svmatthews issue.  When I would attempt to access `http://localhost:8000/bower_components/turf/turf.min.js`, I would get a 404.

To speak more on this topic, `goat` seems to work but it seems to be lacking a `--reload` option to reboot the server on code changes.  It's an improvement over the python server but I think #9 would still be advantageous.

knownasilya:
@alukach I'm thinking of adding a `--live-reload` option to `goat`, basically would serve a live-reload script. You'd have to add the script src to index.html though.

knownasilya:
Although I'm a bit on the side of this being a none dev implementation since it uses `npm start`. I'd want to see live-reload only for those developing this, which means they'd be using grunt.

alukach:
I'm not super familiar with JS projects but I think the oddities are coming around how the project is structured.

```
app/
  - css/
    - app.css
  - js/
    - app.js
  - index.html
bower_components/
  - jquery/
  - ...
````

Something feels a bit odd about this.  `index.html` points to the `bower_components` assets as if they live within it's parent directory:

```html
  <script src="../bower_components/jquery/dist/jquery.js"></script>
  <script src="../bower_components/turf/turf.min.js"></script>
  <script src="../bower_components/mapbox.js/mapbox.js"></script>
```

This works if `index.html` is at the path root, where `../bower_components == ./bower_components`.  I think the correct solution (for now) is to instruct `bower` to install all of it's dependencies into `app/static/bower_components`.  This way only a single directory needs to be served. 

See https://github.com/knownasilya/drop-n-chop/pull/2

alukach:
>  I'd want to see live-reload only for those developing this, which means they'd be using grunt.

I'm a bit confused by this.  I would imagine that the only people running `npm start` would be devs. What is the envisioned deployment of this project?  I assumed it would be hosted somewhere, allowing anyone with an internet connection to use it (rather than requiring them to download/run locally).

knownasilya:
There are really two options:

1. Serve this as static data in nginx directly
2. Serve this via a node server (goat) and proxy via nginx

Usually node related web apps go with route 2.

alukach:
Why stop there?

1. Serve this as static data in nginx directly
2. Serve this via a node server (goat) and proxy via nginx
3. Serve via Github Pages
4. Serve via Amazon S3.

Seeing as this is all static, I would push for 3 or 4, as they involve less moving parts.  I don't really think of this project as being node related, rather it uses node to help generate the static content (similar to Ruby's involvement in any Jekyll site).

knownasilya:
True (plus a dozen other ways).

Live Reload has it's place, but that should probably be addressed in a different PR.

svmatthews:
Thanks for the conversation @alukach and @knownasilya! It seems best at this point to figure out how devs can run their own local server easily and focusing on how to deploy later once there is a fully-functioning alpha version. The plan is to have this running as a web-app and not for people to download themselves (that's kind of the reason why the project started :smile:). 

I'm a fan of `node_components` existing within the `/app` directory right now - seems to make the most sense to me. How about creating a new issue in regards to how to serve this in production?

svmatthews:
Working great after trying the new updates on my end! Looks like an update to the `.gitignore` bower components is all that is left.

knownasilya:
@svmatthews all set

alukach:
Not to be overly concerned with details, but changing gitignore from `bower_components/` to `app/static/bower_components` is actually more specific. If left at `bower_components/` [any directory named `bower_components/` would be ignored](http://stackoverflow.com/questions/1470572/gitignore-ignore-any-bin-directory) (not just that at the project root).  Is this the intention?

svmatthews:
Thanks @knownasilya & @alukach!

PR #18, opened 2015-04-01
Add 'grunt serve' tooling to watch while serving

alukach:
By running `grunt serve`, a developer can start a server that runs `grunt watch` simultaneously with `grunt connect`.  Additionally, when `grunt serve` begins, a browser will open automatically.  Browser will reload on file changes.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -35,15 +35,17 @@ module.exports = function(grunt) {
     },
     watch: {
       all: {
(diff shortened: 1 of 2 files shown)


svmatthews:
@alukach has this been added to #19 at all? Which would you like me to merge first?

alukach:
@svmatthews At first #19 was built off of #18 but I did some rebase magic on #19 so hopefully they should be entirely separate and should play well with each other.  No particular order should be necessary.

Issue #8, opened 2015-03-23
incorporate browserify?


mapsam:
@thebigspoon I'm curious if you'd want to use browserify here or stick with using the Grunt [`wiredep`](https://github.com/stephenplusplus/grunt-wiredep) for loading in all of our bower components?

svmatthews:
After #19 we are removing bower from the equation and going to roll with some quick CDNs and begin working on incorporating browserify for module use.

svmatthews:
This is coming up as an issue as people begin looking at #11

alukach:
What does @aaronr think about this?

Why was it that we decided to table browserify in the first place?  I don't quite recall.  I do like the idea of writing this all node-like in the event that we ever want to make an offline-friendly downloadable DNC.

thebigspoon:
@svmatthews @alukach @aaronr 
I think the easiest solution here is just use `browserify` to require the third-party modules you need to require and then expose/tag them to the `L.DNC` namespace. You'll just have to ensure that the browserify build happens at the end so the `L.DNC` namespace is created.

```javascript
var shape_package = require('shapefile-lib-thing');

L.DNC.ShapeUtils = shape_package; 
```

PR #19, opened 2015-04-01
Reorganize project

alukach:
In favor of allowing the project's root to be servable (such as when using `gh-pages`), `index.html` has been moved into the project root.  All compiled data (from `/src`) is created in (and served from) `/dist`. Additionally, `bower_components` are now stored in the repo.


alukach:
The number of files added is a bit absurd.  It may be better to only store uglified/concatenated versions of `bower_components` in `/dist` and keep `bower_components` with `/src` but ignored from the repo.

thebigspoon:
My comments below are not particularly about this pull request. I do want to point out some cumbersome issues about this project structure and it's dependencies. This seems like a relevant place to list things. Hopefully we can collectively come to a decision on these topics and move forward:

1. `jQuery` is not a dependency for this project, which is great, so let's not pretend it is and purge it

2. We are using two NodeJS package managers. Why? Both `mapbox.js` and `turf.js` are in `npm`. The only `bower` dependent thing is wiredep, which I'm not sure is bringing a lot to the table. If we remove `wiredep` and use a different "build-dependency-thingy" tool then we can remove `bower` altogether. Thoughts?

3. If 1 and 2 happen above we can flatten this structure more and use @alukach suggestions above minus the `bower_components` stuff.

alukach:
> jQuery is not a dependency for this project, which is great, so let's not pretend it is and purge it

:+1: Complete agreement.

> We are using two NodeJS package managers. Why? Both mapbox.js and turf.js are in npm. The only bower dependent thing is wiredep, which I'm not sure is bringing a lot to the table. If we remove wiredep and use a different "build-dependency-thingy" tool then we can remove bower altogether. Since browserify is a dependency in the package.json let's just use that. Thoughts?

I've always looked towards Bower on front-end because of its flat dependency tree. Conversely, I've looked towards NPM on the backend because its offerings seem more comprehensive.  In this case, our "backend" is basically just Grunt dev tools and our front-end requires only 2 libraries. Using NPM as our package manager for the front-end would require the project to either a) store a bunch of dev-tools in the repo; or b) not store our front-end requirements in the repo.  A big part of this reorg is to make it deployable by `gh-pages`, meaning that b) isn't much of an option.

I'd suggest we actually drop Bower until it becomes necessary, use NPM for backend dev tools (as we do now, omitting them from the repo), and do one of three options:

* Avoid front-end dependency worries by importing Mapbox.js and Turf.js from CDNs. This makes offline dev a bit more annoying (if 
* Store the dependencies in our repo by hand, as is done with [geojson.io](https://github.com/mapbox/geojson.io/tree/gh-pages/lib). This seems like it would be a hassle but at least we could 
* Manage library versioning with NPM, but create some pipeline via Grunt to uglify / concat front-end JS libraries, storing the product of those in the repo under `/dist`

I vote for the first option today, the third option in the future.

thebigspoon:
> Using NPM as our package manager for the front-end today would require the project to either a) store a bunch of dev-tools in the repo; or b) not store our front-end requirements in the repo

Sorry, I was tired when I wrote that ;-) I agree with your three choices. `turf` and `mapbox` should be from CDN for now. Later we can have `npm` pull and build them dynamically

alukach:
@thebigspoon @svmatthews Okay, I think that should do it.

svmatthews:
:white_check_mark: CDNs for now, `npm` later

:white_check_mark: browserify, already have an issue on this #8 

:question: As far as the workflow for `gh-pages` I'd like to keep the `master` branch and only push pre-defined versions into `gh-pages` unless we're making quick bug fixes. That way we'll always be confident everything going from `master` to `gh-pages` is accurate and will be running along a one-way street, without ever having to merge from `gh-pages` into `master.

PR #20, opened 2015-04-01
fixing connect to run on ./ instead of /app

svmatthews:
@alukach #18 did almost all of it, but updating the `Gruntfile.js` to serve at the base directory because of the updates from #19

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -43,7 +43,7 @@ module.exports = function(grunt) {
     connect: {
       server: {
         options: {
-          base: 'app',
+          base: './',
           livereload: true,
           open: true,
         }


alukach:
Good catch, thanks @svmatthews.

alukach:
This was a bit of a side-effect from having the two PRs be completely separate from each other.  I never know whether it's better to build PRs off one another or to do them separately. On one hand, separate allows the project to pick and choose which to implement more freely, on the other hand it can lead to problems like this where one PR can render the other PR to be incorrect.

svmatthews:
Yah, that's something I've struggled with as well. I think the "separate" is what happens naturally, especially when a project is open to accepting PRs - so I think it's up to the interested parties to make sure they are up-to-date with what they are merging into, and up to the maintainers to be clear when making those decisions.

Issue #5, opened 2015-03-22
handle FeatureCollections


svmatthews:
Right now feature collections are a bit iffy. The `ops.geom.union()` function, for example, tests to see if a passed object has `object.features`, which means it's a feature collection and then uses the first feature `object.features[0]` to do the operation. This doesn't seem very good!

svmatthews:
I wonder if we can ensure that ALL features are converted to feature collections on upload and handle them this way? As long as everything is formatted the same, then we don't have to build `if` checks on them.

svmatthews:
Using [`turf.featureCollection(feature)`](http://turfjs.org/static/docs/module-turf_featurecollection.html) should make this pretty :cake:

morganherlocker:
Check out geojson-normalize and geojson-flatten. Running everything through both of these tends to smooth things out with highly variable input.

```js
geojson = normalize(flatten(geojson))
```

svmatthews:
Oh! Just to be clear, are you saying use the `npm` packages for these? [geojson-normalize](https://www.npmjs.com/package/geojson-normalize) & [geojson-flatten](https://www.npmjs.com/package/geojson-flatten)

Those seem super helpful. Presumably the best way to go about doing this would be getting #8 moving forward so accessing these and future modules will be a breeze.

svmatthews:
@alukach it looks like this will be a good next issue to resolve before we add all of the Turf operations. How do you propose we incorporate other modules if we're not using browserify any longer? Maybe we should incorporate browserify for things like these?

svmatthews:
Does it make sense to have *every layer* inside a single feature collection? That would allow us to simplify the selection system into specific geometries.

I was hesitant about this approach because it makes handling features with multiple polygons difficult, but now that I look into the geojson spec a bit more I realize that this is a non-issue since `geometry` is it's own object class within a `feature` and therefore contains `multiPolygons` as if we had a feature collection.

Therefore, does the entire `layerlist` become a representation of EACH `feature` within the global `featureCollection`? I'd like to see `layerlist` respond to an updated map `featureCollection`, rather than update the map from the `layerlist`.

svmatthews:
Hm, now that I think about it (again) I'm realizing that the above would require ALL leaflet layers to be mushed into a single layer, which doesn't seem good. I wonder if we can convert all `featureCollections` into single `feature`s with `multiGeometry` attributes instead.

nbolten:
[...]

PR #21, opened 2015-04-03
Browserify Rewrite for Issue #8

thebigspoon:
I reworked the existing code into modules and wired them together with `browserify`. It does not matter to me whether this PR is merged or not. I did it as a point of discussion for #8 and to really see where the coupling was between the different 'components' in the application.

### Pain Points
1. There's a lot of coupling between the map, the file-layer thing and the geoprocessing tools. If someone wanted to implement the `convex` turf function they'd spend time editing a couple files and trying to figure out where the global things like `numLayers` and `addLayer` come from like [here](https://github.com/cugos/drop-n-chop/blob/master/src/js/_operations.js#L11). Something doesn't feel right about having the `parser` thing that reads the file have a reference to a `map` thing that adds the layer like it does [here](https://github.com/cugos/drop-n-chop/blob/master/src/js/_fileReader.js#L12). That might just be me though. 

2. The existing structure also makes it harder to test in headless NodeJS testing environments like `jsdom` or `phantom`. I like `browserify` for this reason, but maybe it's not what people want or like to use.

### Grunt Problems and Browserify Build CMD
The `grunt-browserify` plugin seems old and runs an infinite loop that nerver terminates. I'm not a huge fan of `grunt` so not sure if i was doing something wrong. But since `npm` can execute commands the browserify build command is `npm run browserify`

### If Not Browserify Then What?
We don't have to use `browserify`. There's other options. But if we want to keep things organized and we expect the application components will follow more of an observable pattern in the future with events then build order becomes important. I'd be open to doing something along the lines of Leaflet with `jake` and `deps` as shown [here](https://github.com/Leaflet/Leaflet/blob/master/build/deps.js)


svmatthews:
Wow. This is big and excites me very much. I'm definitely pro browserify, hence the issue being created. It also allows us to bring in other npm module dependencies really simply - like `geojson-flatten` mentioned in #5.

As far as the structure of the js I like it so far. Does every file have to be a `directory/index.js`? What if we want to add other files to the `map/` directory - would that start getting confusing with something being named index? (just a thought)

Is there a need to uglify the `dnc.js` files at all? It would be nice to make sure files in `/dist` are never edited. Wonder how we can run grunt first, then browserify, then Grunt again. Hm...

I like your comments here, especially the TODO's in the files themselves. I really like the plugin idea for the geoprocessing pieces, especially with their required HTML forms. Let's talk about how to structure this in person next week.

Thanks, @thebigspoon! I'm going to wait to merge until I have some time to test and get acquainted with it all.

thebigspoon:
> As far as the structure of the js I like it so far. Does every file have to be a `directory/index.js`? What if we want to add other files to the `map/` directory - would that start getting confusing with something being named index? (just a thought)
We should definitely break these things up into smaller modules and then `index.js` will just `require` them into the "class". It will make everything easier to load into our comprehension RAM and also test. The other question that your comment seems to hint at is what ARE the components that make up this application? That's for us to explore. I guessed and saw three "component-things" interacting -- filelist ( which I called dropzone and should be changed ), the geoprocessing tools and the map. But to really define what those components are that shoudl be isolated from each other we'd need to sit down and talk about the structure of it when you roll into town

> Is there a need to uglify the `dnc.js` files at all? I wonder how we can run grunt first, then browserify, then Grunt again. Hm...
Yes, we will need to run `uglify` across it. We should also have `dev` and `prod` builds where the `dev` is not uglified because that would be horrible to debug.

> Let's talk about how to structure this in person next week. I'm going to wait to merge until I have some time to test and get acquainted with it all.
Yep, lots to talk about. Waiting is good until you are comfortable with any reorg. I'm going to keep pushing and integrating changes into this branch however

svmatthews:
> Yes, we will need to run uglify across it. We should also have dev and prod builds where the dev is not uglified because that would be horrible to debug.

Can we make a source map when compiling with browserify? I think it's fine to keep the `.js.map` file in the `prod` for now since we're not too close to pushing a real production version yet. Looks like browserify recommends using [excorcist](https://github.com/thlorenz/exorcist)

> Grunt Problems

I wonder if moving this into **gulp** would make things a bit easier? That let's us write purely in javascript and we can make sure browserify and the uglification work hand-in-hand nicely. Just a thought.

dist/js/dnc.js
```diff
@@ -0,0 +1,398 @@
...
+
+};
+
+Map.prototype = {
+
+    init : function() {
+
+        // wireup the other classes 
+        DNC.dropzone = new Dropzone()
+        DNC.menu = new GeoMenu()
```
alukach: I'm not super sharp on auto semicolon insertion. Is this valid?

On a style level, should it be allowed?

thebigspoon: @alukach 
I think "syntax validity" is determined largely by us and what we configure the linter to throw errors about. So there's work to be done there.

thebigspoon:
Since we aren't going to merge this anytime soon, I'm going to close this PR. 

I think there's a cleaner and more succinct solution anyway. Let's think about reusing as much of the existing Leaflet API for DOM manipulation, event registration/firing etc and control inheritance as possible. It makes the most sense to me at least.

There's, still quite a bit of work to do, but [this commit shows](https://github.com/cugos/drop-n-chop/commit/cdf433433295d11b31cd634bde7caa060754feb3) how we'd break out the DropZone class as it's own handler

svmatthews:
@thebigspoon where's the best spot for me to work? Would like to start building a notification UI component a la #25

thebigspoon:
@svmatthews 
not sure, I'd say just work off of master. We can always fold it in later when the structure is mostly baked. Things are in high flux right now. Sorry about that ;-)

PR #24, opened 2015-04-04
Support Browserify within Grunt

alukach:
Add Browserify support.

Additionally break up `watch` tooling between different file types.


alukach:
`grunt serve` lives!

PR #27, opened 2015-04-11
Let's use Leaflet as a Framework for our Application

thebigspoon:


Gruntfile.js
```diff
@@ -48,25 +52,40 @@ module.exports = function(grunt) {
...
       }
+    },
+    concat: {
+        basic: {
+          src: ['src/js/DncApp.js'],
+          dest: 'dist/js/dnc.js'
+        },
+        extras: {
+          src: ['src/js/L.DNC.js', 'src/js/**/*.js'],
+          dest: 'dist/js/L.DNC.js'
```
alukach: I'm a bit confused by this.  What's the difference between `dnc.js` and `L.DNC.js`?

thebigspoon: oh, looks like we forgot to remove it when we cleaned up everything. Now it's all Leaflet. Only this is needed now:

```bash
+        basic: {
+          src: ['src/js/L.DNC.js', 'src/js/**/*.js'],
+          dest: 'dist/js/L.DNC.js'
+        },
```
let me fix that quick like

thebigspoon: @alukach go ahead and pull master it should be fixed

PR #29, opened 2015-04-12
Update gruntfile

alukach:
Running `grunt` now lints, builds, tests the project, starts a server, and opens up a browser.  Any subsequent changes to `index.html` or files within `src/`  re-runs related lint, build, and test operations and reloads the browser window.

If one wants to only build, they can run `grunt build`
If one wants to run the server but skip the primary build, they can run `grunt serve` (this will still build/reload-browser on file changes)
If one wants to run the server without watching the files, they can run `grunt connect`


svmatthews:
after an `npm install` all seems good on my end!

PR #30, opened 2015-04-12
Menu Refactor

alukach:
I've opened this up for review/discussion.  It was built off of the `feature/grunt_helpers` branch (#29), so for sake of a cleaner diff I've set this to merge into that branch.

## What's changed

In efforts to prepare for more menu options, I've broken all of the geo-tooling out of `L.DNC.Menu` and placed it in a `L.DNC.GeoMenu` object that inherits from `L.DNC.Menu`.  The menu now builds itself based on it's own `buildMenu()` function.


dist/js/L.DNC.js
```diff
@@ -387,30 +456,11 @@ L.DNC.MapView = L.Class.extend({
...
-
-        // override defaults with passed options
-        L.setOptions(this, options);
-
-        this._jsonLayerList = jsonLayerList;
-        this.addEventHandlers();
-
-    },
+L.DNC.GeoMenu = L.DNC.Menu.extend({
+    title: "Geoprocessing Tools",
```
alukach: Should this go into `this.options`?  Probably...

dist/js/L.DNC.js
```diff
@@ -480,6 +505,27 @@ L.DNC.Menu = L.Class.extend({
...
+
+        // Build menu dropdown
+        html = '<div class="menu-dropdown menu-expand">';
+        for (var i=0; i < keys.length; i++) {
+            var key = keys[i];
+            html += '<button class="menu-button menu-button-action" id="' + key + '">' + key + '</button>';
+        }
+        html += '</div>';
+        domElement.appendChild(this._buildElement(html));
+        return domElement;
```
alukach: As all menu classes will do something similar to this, this should probably be generalized and placed in `L.DNC.Menu`

alukach:
Thinking more about this, I think this still has quite a ways to go.

There should probably be a single location (likely within `src/js/menu`) that constructs the menu.  The menu will be composed of something like:

```javascript
L.DNC.MenuBar = L.Class.extend({
    initialize : function( options ) {
        new L.DNC.Menu("Geoprocessing Tools", { 
          operations: [
            // The meat of the tooling can either be defined here or in a class that subclasses L.DNC.MenuOperartion
            new L.DNC.MenuOperation("Buffer", function() {
              // Do some buffering stuff
            }),
            new L.DNC.MenuOperation("Erase", new L.DNC.EraseOperation({}))
          ]
        });
        new L.DNC.Menu("Save Data", {
          operations: [
            // ...
          ]
        })
        
    }
});
```

or maybe more like:

```javascript
L.DNC.MenuBar = L.Class.extend({
    initialize : function( options ) {
        new L.DNC.Menu("Geoprocessing Tools", {})
          // These operations would all extend the same MenuOperation (and maybe GeoOperation) class
          .addOperation(new L.DNC.BufferOperation({}))
          .addOperation(new L.DNC.EraseOperation({}))
          .addOperation(new L.DNC.UnionOperation({}))
        ;

        new L.DNC.Menu("Save Data", {})
          // These operations would all extend the same MenuOperation (and maybe GeoOperation) class
          .addOperation(new L.DNC.DownloadJsonOperation({}))
          .addOperation(new L.DNC.DownloadShapeFileOperation({}))
          .addOperation(new L.DNC.CreateGistOperation({}))
        ;
    }
});
```

I'm thinking probably the latter.

To reiterate from the last conversation, a menu operation should be aware of:

- Types Supported (Points, Polygons, FeatureCollection, etc)
- Number of args of supported
- Menu Settings
  - ~~Which menu it belongs in (_can we support submenus?_)~~ (This will be controlled within the menu definition. Regarding submenu support, we should ensure that `addOperation` supports being given a `L.DNC.Menu` instance)
  - How it looks in menu (HTML)
- Dialog Settings
  - Form components
  - Reorder Layers

alukach:
Continuing with this, I think abstracting all Turf operations would be even more ideal:

L.DNC.MenuBar = L.Class.extend({
    initialize : function( options ) {
        new L.DNC.Menu("Geoprocessing Tools", {})
          // These operations would all extend the same MenuOperation (and maybe GeoOperation) class
          .addOperation(new L.DNC.TurfOperation("buffer", {
            maxFeatures: 1
          }))
          .addOperation(new L.DNC.TurfOperation("erase", {
            minFeatures: 2,
            maxFeatures: 2,
            orderImportant: true
          }))
          .addOperation(new L.DNC.TurfOperation("union", {
            maxFeatures: 2,
            minFeatures: 2
          }))
        ;

        new L.DNC.Menu("Save Data", {position: "right", icon: "fa-floppy-o"})  // Pass in menu options
          .addOperation(new L.DNC.DownloadJsonOperation({}))
          .addOperation(new L.DNC.DownloadShapeFileOperation({}))
          .addOperation(new L.DNC.CreateGistOperation({}))
        ;
    }
});

Turf Operations Options
- `supportedFeatures` 
  - Types Supported (Points, Polygons, FeatureCollection, etc)
  - `Array`, default to all
- `minFeatures`
  - Minimum number of layers required
  - `Number`, default to 1
- `maxFeatures`
  - Maximum number of layers required
  - `Number`, default to 0
- `orderImportant`
  - Whether order of layers is important. If `true`, enables reordering in dialog
  - `Boolean`, defautl to `false`
  
Experimental
It seems that while some operations (ex `buffer`) can only be applied to a single layer, there's no reason that we couldn't enable running the same buffer operation on any number of selected features.

- `iterateSafe`
  - Whether it's save to iterate over a list of features (in no particular order) and apply operation on each

Abstracting every turf operation may be lofty, but if we can do it correctly it could allow us to quickly support all operations with little code.

svmatthews:
This is really cool - I'm super into the last example that you're showing. The structure seems really straightforward and easy to manage. It will be particularly helpful as we make decisions on **what** goes into each menu and need to change order/location quickly. Doesn't involve going to individual files for each of the processes.

src/js/menu/operations/TurfOperation.js
```diff
@@ -0,0 +1,49 @@
...
+        minFeatures : 1,
+        maxFeatures : 0,
+        orderImport : false,
+        // iterable : true
+        additionalArgs: null // Kludge to handle no dialog for input
+    },
+
+    initialize: function ( title, options ) {
+        L.DNC.Operation.prototype.initialize.call(this, title, options);
+    },
```
alukach: No need to override method to simply call parent method.

src/js/menu/operations/Operation.js
```diff
@@ -0,0 +1,36 @@
+L.DNC.Operation = L.Class.extend({
+
+    options: {
+        // supportedFeatures : [],
+        minFeatures : 1,
+        maxFeatures : 0,
+        orderImport : false,
+        // iterable : true
```
alukach: These probably only belong in `L.DNC.TurfOperation`

src/js/menu/operations/Operation.js
```diff
@@ -0,0 +1,36 @@
...
+        this._addEventHandlers();
+    },
+
+    _addEventHandlers : function () {
+        this.domElement.addEventListener('click', function(){
+            this._execute.call( this );
+        }.bind(this));
+    },
+
+    // Append button within provided element
```
alukach: Update note

svmatthews:
I wonder if we could also add an "info" parameter that allows us to write a paragraph about the specific type of operation. I.e. when you hover on `buffer` for long enough the `alt` comes up with the options and what a `buffer` actually is. Could just be passed as an HTML string for now.

alukach:
> I wonder if we could also add an "info" parameter that allows us to write a paragraph about the specific type of operation.

@svmatthews That's funny, I was thinking about the exact same thing today.  I think I'm going to take up tooltips as the next feature.

For now, I think that this is complete.  I've met the minimum `buffer`, `union`, and `erase` functionality that was previously wired together.  It should be a breeze to add other Turf operations.  With that said, there's still some work ahead. Notably, figuring out a way to create a dialog based on the TurfOperation options and to wire that up when needed (the `buffer` size use-case is a good place to start).

I also went ahead and moved that event handler into the FileReader logic that you mentioned in #31. It wasn't as big of a deal as I originally thought.

This stuff probably is worthy of creating some tests, although I haven't gotten to that just yet.

svmatthews:
Oh man this is amazing. This whole pub/sub system seems really great. I'm going to work around on this and will merge it asap, then rebase for #33 to start doing notifications based on events.

svmatthews:
:boom: shakalaka!

svmatthews:
When these refactors happen, I'm going to start making wiki pages so we can document once they are settled. Here's the placeholder for the menu architecture. https://github.com/cugos/drop-n-chop/wiki/Menu-Items

PR #31, opened 2015-04-12
Reorg to place main component wiring in L.DNC.js

alukach:
Previously, a huge chunk of the app's initialization took place in `L.DNC.MapView._registerEventHandlers()`, which felt a bit strange.  In effort to keep things modular and avoid interdependency, this wiring was moved in `L.DNC.init`, which seems to be an appropriate place to tie everything together.

### DIFF
--- a/src/js/L.DNC.js
+++ b/src/js/L.DNC.js
@@ -11,7 +11,22 @@
         window.L.DNC = DNC;
 
         L.DNC.init = function(){
-            L.DNC.mapView = new L.DNC.MapView();
+
+            this.mapView = new L.DNC.MapView();
+            this.dropzone = new L.DNC.DropZone( this.mapView._map, {} );
+            this.layerlist = new L.DNC.LayerList( { layerContainerId: 'dropzone' } ).addTo( this.mapView._map );
+            this.geoMenu = new L.DNC.Menu( this.layerlist, {} );
+
+            // examples of events that L.DNC.DropZone.FileReader throws
+            this.dropzone.fileReader.on( "fileparsed", function(e){
+                // TODO: This should be refactored so that this.dropzone and
+                // this.layerlist are not so tightly coupled. The logic behind
+                // this tooling should exist within their respective modules.
+                console.debug( "[ FILEREADER ]: file parsed > ", e.file );
+                var mapLayer = L.mapbox.featureLayer(e.file);
+                this.layerlist.addLayerToList( mapLayer, e.fileInfo.name, true );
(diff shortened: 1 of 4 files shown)


svmatthews:
I agree with this, but it feels a bit weird to have the filereader event handlers in the `init` function though. What if we brought over the first parts like you're doing with the plugins:

```javascript
// wire L.DNC plugins
this.dropzone = new L.DNC.DropZone( this._map, {} );
this.layerlist = new L.DNC.LayerList( { layerContainerId: 'dropzone' }).addTo( this._map );
this.geoMenu = new L.DNC.Menu( this.layerlist, {} );
```

but kept the filereader event handlers `this.dropzone.fileReader.on( "fileparsed", function(e){ ...` within `DropZone.FileReader.js`

src/js/L.DNC.js
```diff
@@ -11,7 +11,22 @@
...
+            // examples of events that L.DNC.DropZone.FileReader throws
+            this.dropzone.fileReader.on( "fileparsed", function(e){
+                // TODO: This should be refactored so that this.dropzone and
+                // this.layerlist are not so tightly coupled. The logic behind
+                // this tooling should exist within their respective modules.
+                console.debug( "[ FILEREADER ]: file parsed > ", e.file );
+                var mapLayer = L.mapbox.featureLayer(e.file);
+                this.layerlist.addLayerToList( mapLayer, e.fileInfo.name, true );
+                this.numLayers++;
+            }.bind(this));
```
alukach: @svmatthews I agree that putting this `fileReader` event handler into `init` feels wrong.  I originally pulled it out of `L.DNC.MapView` which felt more wrong.  The reason that I didn't move it into the `fileReader` (as was done with the `this.dropzone.fileReader.on( "enabled",...` handler that was in `MapView`) was that the handler currently requires access to `L.DNC.layerlist` and `L.DNC.numLayers`.  Rather than embed tightly coupled code deep into `dropzone`, I put the `TODO` on line 22 and figured we'd resolve it later through some better event handling (which is something that I don't really know much about ATM).

I could move it, but worried that it would be "out of sight, out of mine" in the `fileReader` module rather than an obvious need for refactoring within `init`.  Feel free to make the call, it's kind of six vs half dozen.

svmatthews:
Definitely. I think this is where working within the Leaflet spectrum will come in handy. Sounds like we can build the pub/sub listeners within each module to act on other modules, which will completely solve what you're talking about. For now this should work well.

PR #33, opened 2015-04-13
creating Notifications.js to prototype alert system

svmatthews:
This branch adds a new notification platform under `src/notifications/Notifications.js` that allows us to specify "alerts" across the code when we see it being useful for the user. A notification is set up to send a `text` and `type` parameter. `text` can be anything, and `type` styles are defined in the `_notifications.scss` file. Currently we have `.alert` and `.success`

### TODO

- [x] Determine the best way to handle notifications. ~~1) currently as-is, by sending notes from anywhere within the source code, OR~~ 2) by building our pub/sub listeners to relay events when needed. This way we could handle all events within the `Notifications.js` file.
- [ ] add notification tests
- [ ] ~~**option**: add notification "locations" other than top right~~
- [x] **option**: specify the time a notification should last on the screen
- [x] style the default notification (probably white, with blue text, and a :hocho: as the icon)
- [ ] ~~**option**: allow developer to write custom notification? I.e. custom icon/color/size~~

### New Ideas

* It seems like it would be cool to create a "history" object that holds ALL use interactions and geometry instances. This would allow us to create an "undo" button, for instance.


svmatthews:
Here's a screen cap of the notifications in action.

dnc-notifications

powersa:
:+1:

svmatthews:
@thebigspoon or @alukach I'm having trouble figuring out the pub/sub system. I see that the FileReader is firing a `fileparsed` event, but am unsure how to pick that up within `Notifications.js` without manually adding the the `fileReader` object into the notification scope. Are we supposed to be sending events into some sort of globally accessible piece, such as `map` that we plan to access all published events from?

Right now I have `this._fileReader = fileReader;` but imagine something like `this._events = events` so we can access anything for the Notifications instead of having to add new objects to the `initialize` functions.

svmatthews:
@thebigspoon how should I write tests for these?

svmatthews:
This is ready to go, except for tests (I know, I know) but I don't know how to write them yet. SUPER basic notification system that we can build off of as we see needs.

alukach:
@svmatthews Sorry, I've been entirely silent on this PR. This may be a bit late, but regarding your comments:

> Currently we have .alert and .success

Great, glad to hear that we have a BAD/GOOD differentiation. 

> do you write three unique notifications or one with the file names comma separated?

I think three unique notifications is cleaner.

> I'm having trouble figuring out the pub/sub system

I don't really know much about the right/best-practice way to follow this pattern in JS. Did you get that figured out?  Maybe you can show me (or I can pick it up while reading your code).

> how should I write tests for these?

I feel pretty good about testing with Karma / Mocha now after writing [the tests for L.DNC.Menu](https://github.com/cugos/drop-n-chop/blob/7efbc44f4de6ffc8dda3085deccc999abd5a096c/spec/suites/L.DNC.Menu.js). Take a look at those. If you'd like, we can do a screen share and go over how I typically write tests for code and get some tests started.

Looks like a great feature!

svmatthews:
Okay, going to pull the :gun: on this one. Test to come l8r.

src/js/mapview/MapView.js
```diff
@@ -25,7 +25,7 @@ L.DNC.MapView = L.Class.extend({
     _setupMap : function () {
 
         L.mapbox.accessToken = 'pk.eyJ1Ijoic3ZtYXR0aGV3cyIsImEiOiJVMUlUR0xrIn0.NweS_AttjswtN5wRuWCSNA';
-        this._map = L.mapbox.map('map', 'svmatthews.hf8pfph5', {
+        this._map = L.mapbox.map('map', 'examples.map-zr0njcqy', {
```
alukach: @svmatthews Noticed that my basemap now says "EXAMPLE" and comes with some points of interest around the DB area.  Is this intentional?
screen shot 2015-04-29 at 9 43 34 pm

svmatthews:
Yah, it was sucking down my map views :wink: but now that we got a ton of views from maptime I'll change that back up!

Issue #32, opened 2015-04-13
/dist conflicts ALL THE TIME


svmatthews:
Is it just me or do we keep getting stupid branch conflicts because of the `/dist` builds changing all the time? Wondering if it's worth removing `/dist` in the `.gitignore` ...

alukach:
Total agreement on this.  I think this ties into #28.  My suggestion is as follows:

1. Add `/dist` to `.gitignore`
1. Devs develop on branches off of `master`, no longer worrying about `/dist`
1. A bot watches `master`. When a commit (merge) is made on `master`, the bot checks merges `master` into `gh-pages`, lints, contats, uglifies, and tests `/dist`. If all passes successfully, it force adds `/dist` (`git add ./dist -f`) to the repo, commits, and pushes.

This way, we'd have a continuous deployment system without having to sweat `/dist` conflicts.  In #28 you mentioned that you may not want to be deploying every accepted PR.  In that case, you could create a branch like `production` in that is watch by the build bot in the same way that it was watching `master` in my scenario (although, I still like the idea of deploying all accepted PRs ASAP).

I know little-to-nothing about Travis, but a quick 30 seconds search makes me this that this _may_ be possible.

svmatthews:
I'm not totally set on being strict about merges into `gh-pages` but just want to make sure `master` is our reliable dev branch that people work with rather than individuals accidentally forking the `gh-pages` branch and working off of that. Then we'll run into *real* merge conflicts, rather than stupid ones like this issue :wink:

svmatthews:
Fixed in #34

PR #34, opened 2015-04-14
No longer store dist/ in repo

alukach:
As per #32


svmatthews:
Thanks for doing this! Looks like Travis is running off of `/dist`, which may make testing a bit more complicated. Might need to have @thebigspoon chime in on this one.

alukach:
@svmatthews Okay, this should be good to go now!

svmatthews:
:+1:

svmatthews:
:+1:

PR #26, opened 2015-04-08
Intersect function

hamhands:
### What changed
* Added intersect function

### Screencap
new

PR #35, opened 2015-04-14
add intersect function

hamhands:
### What changed
* Added intersect function to handlers and ops in menu.js
* Implements changes started in #26

### DIFF
--- a/src/js/menu/Menu.js
+++ b/src/js/menu/Menu.js
@@ -81,6 +81,18 @@ L.DNC.Menu = L.Class.extend({
                 )
             );
         }.bind(this));
+
+        var intersect = document.getElementById('intersect');
+        intersect.addEventListener('click', function() {
+            this.ops.execute.call( this, 
+                this.ops.geom.intersect(
+                    this._jsonLayerList.selection.list[0].layer._geojson,
+                    this._jsonLayerList.selection.list[1].layer._geojson,
+                    this._jsonLayerList.selection.list[0].info,
+                    this._jsonLayerList.selection.list[1].info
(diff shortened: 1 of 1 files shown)


svmatthews:
closing #26 in favor of this PR

alukach:
So, I think there's some good news and some bad news related to this PR.

The bad news is that, after the #30 merge, this PR is no longer compatible.  The way that the menus and operations are constructed is now very different (and largely abstracted).  This means that this change will probably not be accepted.

The good news is that it's now easy-peasy to add support for Turf operations.  All you need to do is add a new configuration in the [MenuBar.js](https://github.com/cugos/drop-n-chop/blob/6bcafaf6b17955c0f291b1909835c586a5536525/src/js/menu/MenuBar.js#L15-L18) indicating which Turf operation you're adding and what the requirements for the operation are.  It should now be super easy to add a ton of the Turf.js operations today, and we'll build out the `TurfOperation` class to accommodate those that aren't supported yet.

Is there any point to keep this PR open or we should try again for a third one?

svmatthews:
Oh man you may have just broken @hamhands's :heart: - I was so confident in his addition and totally blanked on the incoming changes. I'll work with him tomorrow on rebasing this PR and incorporating the new updates. No need to close!

hamhands:
It's cool guys, I got this:
tumblr_mnb4mzxm6j1rks8b7o1_500

svmatthews:
luxk2be

svmatthews:
Sorry @hamhands gotta close this in prep for our sprint next week!

PR #36, opened 2015-04-14
Adding in an option, defaulting to on, that zooms the map to fit new layers when added

aaronr:
Hey there folks.  Just checking out the new project... way cool.  While I was in checking it out I thought I would try to add something to get my feet wet.  Simple change to loop through all layers when a new layer is added and come up with a bounds for all the data and zoom the map to it.  This makes adding data much more pleasant  as you instantly see all your data.  It is an option (defaulting to on) that can be tied to a UI element in the future if you wanted to make it an option (since auto zoom in some special cases might not be desired).  In general though, I imagine that the 99% of cases will like having the map auto zoom to an extent to see all data when new layers are added.

Feel free to reject as well... just playing around tonight and thought I would kick the tires :-)

### DIFF
--- a/src/js/layerlist/LayerList.js
+++ b/src/js/layerlist/LayerList.js
@@ -108,7 +108,7 @@ L.DNC.LayerList = L.Control.extend({
             for (var i in this._layers) {
                 bounds.extend(this._layers[i].layer.getBounds());
             }
-            this._map.fitBounds(bounds)
+            this._map.fitBounds(bounds);
         }
 
     },
--- a/src/js/layerlist/LayerList.js
+++ b/src/js/layerlist/LayerList.js
@@ -14,6 +14,7 @@ L.DNC.LayerList = L.Control.extend({
     // defaults
     options: {
         autoZIndex: true,
+        zoomToExtentOnAdd: true,
         layerContainerId: 'dropzone'
     },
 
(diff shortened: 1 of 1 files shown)


alukach:
Agreed that this is a useful feature.  +1 for having it enabled/disabled within options!

Thanks @aaronr 

ps. full disclosure: I never when it's cool to use `for (var i in this._layers)` and when to use `for (var i = 0; i < this._layers.length; i++)`, I always just opt for the latter.

svmatthews:
Nice! Thanks @aaronr this is really great. I'm excited that you brought up the `options` piece. I'd love to see how we can allow a user to turn these on and off within `Menu.js` (and potentially save their options in the cache so they can have their own browser defaults).

@alukach we should discuss what syntax/spec we want to use in regards to things like for/in and for/length. Making an issue for discussion #37

aaronr:
Since that method is using L.stamp to generate the ID for the layer in the internal layer list... it will be a non contiguous (sparse) array and therefore I think you want to use the for/in syntax.  Just my 2 cents and understanding on why you would want to use that over the for/length in this case.

https://github.com/cugos/drop-n-chop/blob/master/src/js/layerlist/LayerList.js#L88

To be honest though I was just following the convention in the rest of the file as I wanted to fit into common style you guys where already using:

https://github.com/cugos/drop-n-chop/blob/master/src/js/layerlist/LayerList.js#L125

aaronr:
Oops... to late to be thinking about this stuff... _layers is an object not an array.  Anyway, seems natural to me to use for/in for objects, and for/length for contiguous arrays.

And with that I head to bed...

Issue #41, opened 2015-04-14
Mv Menu configuration to L.DNC.js


alukach:
Thinking about the design that came out of #30, I feel that instantiating / configuring the menu within `menu/MenuBar.js` was a mistake.  I'm now thinking that all configuration / setup should be done in a centralized place, such as `L.DNC.js`.  This way, hopefully developers should have less hunting to do when figuring out where elements are coming from.

alukach:
Fixed in #43

PR #43, opened 2015-04-15
Move initialization into L.DNC.js, normalize object apis

alukach:
As mentioned in #41, I felt that the menu configuration should be moved into `L.DNC.js`. This allows for all the classes within `menu/` to act as more flexible constructor pieces.  Additionally, I tried to adopt a pattern for the API `addTo`, `addChild`, as was used by the `LayerList` (well, `addTo` at least).

I wanted to get this in quick before we throw @hamhands for another loop.

### DIFF
--- a/src/js/menu/MenuBar.js
+++ b/src/js/menu/MenuBar.js
@@ -1,32 +1,29 @@
 L.DNC = L.DNC || {};
 L.DNC.MenuBar = L.Class.extend({
 
-    // defaults
-    options: {
-        parentId : 'menu-bar'
+    initialize: function ( options ) {
+        L.setOptions(this, options);
+        this.children = [];
+        this.domElement = this._buildDomElement();
     },
 
-    initialize: function ( layerlist, options ) {
-        // override defaults with passed options
-        this.layerlist = layerlist;
-        L.setOptions(this, options);
+    // Add object as child. Object must have domElement property.
+    addChild: function ( child ) {
+        this.domElement.appendChild(child.domElement);
+        child.parentDomElement = this;
+        this.children.push( child );
+        return this;
+    },
 
-        var geotools = new L.DNC.Menu( "Geoprocessing Tools", {} )
-            .addOperation(new L.DNC.TurfOperation('buffer', {
-                maxFeatures: 1,
-                additionalArgs: 0.1
-            }))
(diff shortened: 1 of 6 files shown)


src/sass/_nav.scss
```diff
@@ -8,7 +8,6 @@
   left:250px;
   background:$white;
   padding:5px;
-  height:41px;
```
alukach: I dropped the height in efforts to allow for multiple menu-bars to be possibly added within the `nav#menu-bar` element (this is probably not likely).  I figured it would also help for lower-resolution screens (rather than pushing menus out of view).

Without the `height` styling, the menu bar appears as it did before.

src/js/menu/MenuBar.js
```diff
@@ -1,32 +1,29 @@
...
-            }))
-            .addOperation(new L.DNC.TurfOperation('erase', {
-                minFeatures: 2,
-                maxFeatures: 2
-            }))
-            ;
+    // Append this domElement to a give domElement
+    addTo: function ( parentDomElement ) {
+        parentDomElement.appendChild(this.domElement);
+        this.parentDomElement = parentDomElement;
```
alukach: Every menu object should now have a `parentDomElement` attribute.

src/js/menu/MenuBar.js
```diff
@@ -1,32 +1,29 @@
...
 
-    initialize: function ( layerlist, options ) {
-        // override defaults with passed options
-        this.layerlist = layerlist;
-        L.setOptions(this, options);
+    // Add object as child. Object must have domElement property.
+    addChild: function ( child ) {
+        this.domElement.appendChild(child.domElement);
+        child.parentDomElement = this;
+        this.children.push( child );
```
alukach: Every menu object should now have an array of child attributes at `this.children`.  Each of these children are objects, all of which should have a `domElement` attribute.

svmatthews:
I like that things are getting initialized all in one, but am wondering how large this file is going to get once we have ALL of the Turf operations, menu options, downloads, other processes, etc. Is that going to be an issue @alukach?

alukach:
It could get big, yes, but we could probably figure out a method to split it up.  I say we cross that bridge when we get to it. :bridge_at_night:

svmatthews:
Roger. Let's do it!

Issue #38, opened 2015-04-14
create script that builds from source and pushes into production


svmatthews:
Now that we aren't watching `/dist` from #34 it would be good to get a shell script or some sort of automated way of building from the `master` branch and pushing into `gh-pages` for the production site. This presumably would only work for those who have write access and if they actually ran the script themselves. What does that workflow look like?

alukach:
I've actually got this 90% of the way done: https://github.com/alukach/drop-n-chop/blob/master/.travis.yml  .  The intended feature is to build the project and deploy it to the `gh-pages` branch at the end of every successful Travis build on `master`.

svmatthews:
I had no idea you could run things `after_success` in Travis. That's really neat! Also removes the whole step of one of us needing to run the script manually from our machines. Nice work!

alukach:
Fixed in #44

PR #44, opened 2015-04-15
Add continuous integration tooling

alukach:
After a successful build, Travis will run `.travis/deploy-to-gh-pages.sh`. This script will check to see that the build took place on the `RELEASE_BRANCH` (configured as a variable within the Travis configuration file), and if it has [`grunt gh-pages`](https://github.com/tschaub/grunt-gh-pages) is ran, copying the repo (including ignored files/dirs, excluding dot files/dirs) to the `gh-pages` branch.  This requires that Travis have a [Github OAuth token] added as a variable (this is done as a hidden variable on the Travis website, rather than within the config file).  I've gone ahead and set Travis up with a token.

Currently, `RELEASE_BRANCH` is set as `master`, meaning any commits to `master` will be deployed. If this proves to be a source of error, we can adjust as necessary.

This addresses #38

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,3 +5,7 @@ before_install:
   - gem install sass
 before_script:
   - grunt build:basic
+after_success:
+  - .travis/deploy-to-gh-pages.sh
+env:
+  - RELEASE_BRANCH=master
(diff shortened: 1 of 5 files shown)


svmatthews:
@alukach this is really special. So many freaking things happening I can't keep my head straight but I'm trusting you 100% on this one. Sounds like we should wait until @aaronr finishes up #45 before this merges in. But otherwise :+1: here

alukach:
@svmatthews Do you think so? I feel like this is independent of #45 (or maybe a requirement for #45). This has functionality on its own. I'm also a fan of keeping as few PRs open at any given time, helps make merges a bit easier. Open to merging it?

svmatthews:
:+1:

PR #46, opened 2015-04-16
entire document is dropzone, working on #39

### LINKED ISSUE #39: Bind dropzone to body
Currently, the "dropzone" is only the map area. I have made the mistake of dragging files to the filelist area (which seems like a natural place to drop them to me), which opens the file in my browser rather than loading it into DNC.  By binding the dropzone to the entire body, this would be avoided.

svmatthews:
### What changed?

* updated dropzone container to be `document.body`, resolves #39 
* updated `#fileList` to match `layerlist` phrasing (new file name & #id)

dnc-dropzonefix


alukach:
Looks good!

Issue #45, opened 2015-04-15
Make ability to have grunt build release and dev instances of project


aaronr:
There is no current way (other than modifying a tracked index.html file) to change to use non-minified versions of the code.  Also the index.html file is not placed in the dist dir like it should.  I understand that there where reasons to do this... but there must be a better way:

1) Have index.html end up in dist like the rest of the project when built.

2) Change the gh-pages stuff to just publish the contents of dist.
  - This gets rid of all the extra stuff that is being published that should not be (config files etc)
  - Everything is at the right level then (i.e. index.html is no longer treated special)

3) Make index.html based on a template in the root folder, but use a grunt html processor to build out the real index.html based on build type.  Some possible grunt tools already exist:
  - https://github.com/spatools/grunt-html-build
  - https://github.com/dciccale/grunt-processhtml

4) Make grunt serve and the connect stuff for grunt serve from the dist dir.

In the end you would be able to specify build type... each would pre-process the index.html to the proper form, populate the dist dir, and if updating master push a release version of the dist dir to gh-pages.

I plan to work on prototyping this (as a learning exercise for myself) in the aaronr_gruntwork branch

alukach:
+1 to being able to develop off of `L.DNC.js` and serve off of `L.DNC.min.js`, the uglified file definitely takes some fun out of the dev process.  My $0.02: I think what would make this the easiest/most-natural-to-develop-on would be to have the dev folder hierarchy match what we want the prod folder to be (starting with moving `index.html` into `src/index.html`). Basically, any assets that would make their way into prod should go into `src`. Is this kind of along the lines of what you were thinking?

Also, in the event that you haven't already looked into it, the solution to #2 is super trivial (assuming #44 is merged), you would only need to change the [`base`](https://github.com/cugos/drop-n-chop/pull/44/files#diff-35b4a816e0441e6a375cd925af50752cR73) arg to point to `dist/`.

svmatthews:
:+1: to all of this. Take it away @aaronr!

alukach:
Fixed in #48

PR #48, opened 2015-04-16
Switch to a 'dev' and 'prod' build system in Grunt

aaronr:
This pull request adds two new modules (processhtml and focus) to allow us to more consistently have a dev and prod build.  Now all parts of the task chain have a dev and prod version... and grunt will default to dev.  
Also added a template in index.html so that a dev build will import the standard JS and a prod build will use the minified JS.
This will require a npm install to get the new modules.  
We need to figure out the gh-pages work quickly as this will break our ability to push the master branch directly to the gh-pages branch.  @alukach is working on integrating https://github.com/tschaub/grunt-gh-pages into our project (which can push the dist dir to the gh-pages branch).


aaronr:
This is to address issue #45

alukach:
So, looking at this odd thing where Travis built and deployed to `gh-pages`.

The way our system works is that when a build passes tests, Travis runs the `deploy-to-gh-pages.sh` script.  In the script, [these lines](https://github.com/cugos/drop-n-chop/blob/080ee563546c23a91662e643ac4ea28245e49e68/.travis/deploy-to-gh-pages.sh#L15-L18) are supposed to abort if you're not building/testing the `master` branch.  Those failed.

Re-reading the Travis docs, looks like I was confused by what the `TRAVIS_BRANCH` environment variable represented:

> TRAVIS_BRANCH:For builds not triggered by a pull request this is the name of the branch currently being built; whereas for builds triggered by a pull request this is the name of the branch **targeted by the pull request** (in many cases this will be master).

I assumed it meant that it was the branch being built, not to target pull request.  Will make a quick fix.

README.md
```diff
@@ -37,6 +37,16 @@ All of this work is made possible by [CUGOS](http://cugos.org), an open-source g
 
 **NOTE:** Things are moving very quickly with DNC right now. We are still continuing to build out and refactor the architecture of the application. If you want to read more about our decision-making process take a look at some of our [meeting notes in the wiki](https://github.com/cugos/drop-n-chop/wiki/Meeting-Notes---04-11-2015).
 
+Run a local server with a development (not minified) version of the code:
+
```
alukach: Looks like we're missing the command here

alukach: Actually, maybe these additions to the `README.md` should be rolled into the text on line `36` of the file

alukach:
:+1: :+1: :+1: :+1: :+1: :+1: :+1: 

Kind of a wild ride, that one...

PR #53, opened 2015-04-18
Remove Cross Component References

thebigspoon:
I would appreciate some feedback about the ideas I mention below and some of the solutions the commits tied to this ticket propose.

### The App/Controller Idea
Originally the `MapView` class was called something like `DncApp` and it's general purpose was to glue and setup things:
- setup and bind components being used in DNC, like the map
- setup component event listeners such as the `FileReader` handlers
- it was a place for slim applciation logic, but not component specific stuff
- it was also the hub where components could reference and talk with each other

When we decided that `MapView` felt like it was it's own component, things got shuffled around and some glue stuff didn't make sense to have in that class anymore.

I've noticed that some of the glue logic was moved out into `L.DNC` where it currently lives. My guess is that we can expect that `glue` to grow in the future. `L.DNC` really feels like a simple namespace object for things to hang off, but not it's own class. Maybe that's just me.

More importantly, another side effect of the refactor was the propagation of some patterns that I feel like we should avoid:
- `FileReader` now reaches outside of it's scope globally to access other components [here](https://github.com/cugos/drop-n-chop/blob/master/src/js/dropzone/handlers/DropZone.FileReader.js#L30-L31).
- The `TurfOperation` also reaches outside of it's scope to add layers [here](https://github.com/cugos/drop-n-chop/blob/master/src/js/menu/operations/TurfOperation.js#L30).
- Even more confusing, the `FileReader` component fires an event that it itself listens to [here](https://github.com/cugos/drop-n-chop/blob/master/src/js/dropzone/handlers/DropZone.FileReader.js#L24-L32). 

### Proposal
It feels like we need a thing that does the glueing and setup again. This thing, lets call it a controller, would also hold component references so we don't force components that shouldn't know about each other to tightly reference each other.

### What's Changed?
- I removed all cross references except [one](https://github.com/cugos/drop-n-chop/blob/remove-reflexive-refs/src/js/menu/operations/TurfOperation.js#L14-L23), which I flagged as a possible removal later. 
- For lack of a better term ( i'm open to anything ), I called the controller `AppController`. See the [class here](https://github.com/cugos/drop-n-chop/blob/remove-reflexive-refs/src/js/controller/AppController.js).


svmatthews:
Sorry for the delay on this one, was busy this past weekend! I'm in favor of this layout - though I was in favor of the last one as well so I think my opinion shouldn't hold too much weight here. 

Just to keep my head straight, let me present a **hypothetical scenario**. If I want to add a "fileparsed" publication to a notification system that would subscribe to such events, would I add another line to the `_addEventHandlers` function with the following:

```javascript
_addEventHandlers: function(){
    this.dropzone.fileReader.on( "fileparsed", this._createNotification.bind( this ) );
    // ... more handlers ...
} ,
```

and then add a new private handler function to deal with the notification information/type

```javascript
// new notification handler that passes options into a create() function
_createNotification: function( e ) {
    this.notifications.create( stuff, anotherThing, etc ); 
} ,
```
??

thebigspoon:
@svmatthews -- yep, that's exactly right

svmatthews:
Excellent. I think this is exactly what I was thinking about in my head but wasn't sure how to phrase it all. When we were talking about the pub/sub piece it seemed like there was something missing that ALL publications were sent to and all subscriptions listened to. This seems to properly handle that disconnect.

:+1: from me

src/js/menu/operations/TurfOperation.js
```diff
@@ -10,7 +11,16 @@ L.DNC.TurfOperation = L.DNC.Operation.extend({
...
 
     execute: function () {
-        var layers = L.DNC.layerlist.selection.list;
+        /*
+        **
+        **  TODO: this is the type of referencing
+        **  that feels like it should be avoided.
+        **  it is the only reference left and it cannot
+        **  be factored out until we potentially
+        **  revisit how MenBar, Menu and Operation work together
```
alukach: While we don't have to solve it all right now, can you give some hints in regards to what you're thinking about a refactor between the MenuBar, Menu, and Operation?

package.json
```diff
@@ -10,7 +10,7 @@
   },
   "main": "src/main.js",
   "scripts": {
-    "start": "goat -e index.html -p 8000",
+    "start": "goat -e dist/index.html -p 8000",
```
alukach: While this is a side-topic from the PR, I would suggest dropping this in favor of `grunt serve:dev`.

spec/es5-shims.js
```diff
@@ -0,0 +1,1431 @@
...
+            var hexRegex = /^0[xX]/;
+            return function parseInt(str, radix) {
+                var string = String(str).trim();
+                var defaultedRadix = Number(radix) || (hexRegex.test(string) ? 16 : 10);
+                return origParseInt(string, defaultedRadix);
+            };
+        }(parseInt));
+    }
+
+}));
```
alukach: I added some shim stuff on my current working branch as well (although this probably is covered now with this shim). This makes me wonder, should this be concatenated with the entire app for browsers?

alukach:
> More importantly, another side effect of the refactor was the propagation of some patterns that I feel like we should avoid:
- `FileReader` now reaches outside of it's scope globally to access other components [here](https://github.com/cugos/drop-n-chop/blob/master/src/js/dropzone/handlers/DropZone.FileReader.js#L30-L31).
- The `TurfOperation` also reaches outside of it's scope to add layers [here](https://github.com/cugos/drop-n-chop/blob/master/src/js/menu/operations/TurfOperation.js#L30).
- Even more confusing, the `FileReader` component fires an event that it itself listens to [here](https://github.com/cugos/drop-n-chop/blob/master/src/js/dropzone/handlers/DropZone.FileReader.js#L24-L32). 

Yeah, most/all of these were my bad.  The interlinking of dependencies felt wrong, I was pretty much shooting from the hip.  :+1: regarding these changes.

To further understand, in the scenario described by @svmatthews, the `.on()` tooling comes from `L.Mixin.Events`, correct?

src/js/menu/operations/TurfOperation.js
thebigspoon: I haven't looked at the relationship in depth enough to guess what needs to change now. But in particular it feels like the `L.DNC.MenuBar` and it's inlaws own the selection relationship more than `L.DNC.LayerList` does.

So a quick win in terms of this reference might be to have `L.DNC.LayerList` fire events for add and remove [here](https://github.com/cugos/drop-n-chop/blob/remove-reflexive-refs/src/js/layerlist/LayerList.js#L185-L192) and `L.DNC.MenuBar` and inlaws listen for those and hold the selection internally. 

That's a quick guess shooting from the hip, so not sure if that would work.

package.json
thebigspoon: currently `grunt` jumps into a infite loop when I try to `serve` or `watch`. It's probably just my dev environment, I'll have to figure it out more. But until then let's keep this around as an option.

spec/es5-shims.js
thebigspoon: The shims I added are really only for testing and PhantomJS. In fact, I will create a ticket because there's only one shim in there i really need for `Function.prototype.bind`. So none of these are needed right now in code, but if you see a usecase then maybe we can start putting them in and packaging them

thebigspoon:
@alukach: yep,  `.on()` and `.off()` and all the aliases for them are from `L.Mixin.Events`. 

@alukach and @svmatthews:  when you guys come to a decision about this PR can you merge it in? I'm going to be on a beach for the next week.

svmatthews:
Fine merging this. I think my head is tied a little tighter with this PR and will be able to cruise on some functionality after it is merged!

svmatthews:
@thebigspoon is [this going to be you](https://www.youtube.com/watch?v=6E5m_XtCX3c) next week? You ocean man, you.

package.json
alukach: Cool with me.

spec/es5-shims.js
alukach: Okay.  I think I did see `.map` used in the code once or twice, which I believe is an ECMA5 thing.

alukach:
Pulling the trigger on this one! :gun:

PR #55, opened 2015-04-23
Testing refactor

alukach:
- **Simplified testing suite**: It seemed strange to have `expect.js`, `happen.js`, and `sinon.js` stored in the tests (rather than stored in `package.json`).  Additionally, it seemed that `karma.conf.js` could be simplified by letting the `files` attribute do most of the work.
- **Disabled output on certain tests**
- **Added tests for `Menu`**


spec/suites/L.DNC.DropZone.FileReader.js
```diff
@@ -11,6 +11,7 @@ describe("L.DNC.DropZone.FileReader > ", function () {
         var fileReader = null;
 
         beforeEach(function () {
+            console.debug = function(){};
```
alukach: This should be unmocked after tests.

alukach:
I believe this is ready for merge. Curious to hear what @thebigspoon thinks of these changes since he set up the initial test framework.

spec/karma.conf.js
```diff
@@ -1,44 +1,22 @@
...
-            'karma-chrome-launcher'],
-
         // frameworks to use
-        frameworks: ['mocha'],
+        frameworks: ['mocha', 'chai', 'sinon'],
 
         // list of files / patterns to load in the browser
-        files: files,
+        files: [
+            "node_modules/leaflet/dist/leaflet.js",
```
alukach: I feel like this should be `mapbox.js` as that's what we're actually using in prod.  Differences seem to be minimal, but would prefer to keep tests as close to production as possible.

alukach:
_poke_ :point_right: @svmatthews 

Any thoughts on this? I think that this is at a state to be merged in.  I've been building on top of it in another branch, but want to make sure that this is kosher before I rely on it too much.

thebigspoon:
@alukach nice cleanup work! merge that puppy

svmatthews:
Sorry for the no comment, @alukach. Wanted @thebigspoon to take a look!

PR #56, opened 2015-04-24
Clean up addChild()

alukach:
Two changes:
- **Rely on inheritance for `addChild()`**: Rather than have `L.DNC.Menu.addChild()` duplicate the functionality of its parent's `addChild()` with a small change, it now makes the change and calls the parent's `addChild()`.  Not sure how you guys feel on inheritance-heavy OO design, it can make it difficult to figure out where methods and attributes are coming from, but can prevent a lot of rewritten code.  Maybe we should make `Operation`, `TurfOperation`, `Menu`, and `MenuBars` all inherit for L.Class but pick and choose from a collection of mixins?  This could make things a bit cleaner.
- **Simplify `parent` vs `parentDomElement`**: For some reason, `Operation`, `TurfOperation`, `Menu`, and `MenuBars` were creating attributes called `parentDomElement` that were sometimes a dom elements but were usually objects. Now, when the parent will always be an object if possible.  If the parent is an actual domElement (such as when appending the `MenuBar` to empty `nav#menu-bar`
">`), then the domElement is appended as `parent` (maybe this is wrong? maybe it should be a different attr? I kind of liked it being the same)

### DIFF
--- a/src/js/menu/operations/TurfOperation.js
+++ b/src/js/menu/operations/TurfOperation.js
@@ -46,7 +46,7 @@ L.DNC.TurfOperation = L.DNC.Operation.extend({
         **
         */
         var eventExtras = { mapLayer: mapLayer, layerName: newLayer.name, isOverlay: true };
-        this.parent.parentDomElement.fire('operation-result', eventExtras);
+        this.parent.parent.fire('operation-result', eventExtras);
     },
 
     _validate: function ( layers ) {
--- a/src/js/menu/Menu.js
+++ b/src/js/menu/Menu.js
@@ -10,12 +10,9 @@ L.DNC.Menu = L.DNC.MenuBar.extend({
     },
 
     // Add object as child. Object must have domElement property.
-    addChild: function( child ) {
-        var dropdown = this.domElement.getElementsByClassName('menu-dropdown')[0];
-        dropdown.appendChild( child.domElement );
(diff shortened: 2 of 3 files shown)


src/js/menu/operations/TurfOperation.js
```diff
@@ -46,7 +46,7 @@ L.DNC.TurfOperation = L.DNC.Operation.extend({
         **
         */
         var eventExtras = { mapLayer: mapLayer, layerName: newLayer.name, isOverlay: true };
-        this.parent.parentDomElement.fire('operation-result', eventExtras);
+        this.parent.parent.fire('operation-result', eventExtras);
```
alukach: Can @thebigspoon verify that this is the expected functionality?  Should `.fire` be called on the parent's parent or the parent's dom element?

svmatthews:
@alukach I'm a little confused on this one, but mostly from a lack of understanding of the current system. Pretty ambivalent towards the merge - not sure I'm going to be of use in this one!

alukach:
@svmatthews Any questions in particular? Where do we go from here? I feel pretty good about this PR, would you prefer that we wait to hear from another or is it cool to merge?

svmatthews:
sorry should have been more cleare! :+1: for me on this, but not because I know what's going on ;)

PR #59, opened 2015-04-27
adding function to remove single feature FeatureCollections, #5

svmatthews:
### What changed?

* Adding small helper script when new layers are created to break apart feature collections if they only have a single feature

Here you can see `union` working again since we don't have `FeatureCollections` running through it any longer:
dnc-normalizefix


alukach:
Looks good to me!

PR #63, opened 2015-05-03
Input templates

svmatthews:
This a first attempt at generating forms when the user clicks an operation. My original thought was to use Handlebars.js or Mustache.js to do this, but wanted to see what the burden was of building our own forms. It's not the most beautiful result - but I think it does a good job at continuing to keep all of the information in one place when we register the turf functions.

### What Changed?

* Newly added `Forms.js` under `/menu` - this is a submodule of menu, since most of the interaction with forms will come out of interacting with the menu specifically.
* Publishing of `operation-click` to the `AppController.js` so we can do things when a user clicks those buttons. Right now the only operation that runs when that event fires is the `Form.render()` function which takes in an operation info, i.e. `buffer`.
* New information in the turf menu registration - here's an example:

```javascript
.addChild(new L.DNC.TurfOperation('buffer', {
    maxFeatures: 1,
    additionalArgs: 0.1,
    description: 'Calculates a buffer for input features for a given radius. Units supported are miles, kilometers, and degrees.',
    parameters: [
        {
            name: 'distance',
            description: 'Distance to draw the buffer.',
            type: 'number'
        },
        {
            name: 'unit',
            type: 'select',
            description: '',
            options: ['miles', 'feet', 'kilometers', 'meters', 'degrees'],
            selected: 'miles'
        }
    ]
}))
```

Let me know what you think about this!


svmatthews:
Closing in favor of #67

Issue #57, opened 2015-04-24
CUGOS Spring Fling "alpha"


svmatthews:
The CUGOS Spring Fling is coming up in a few weeks, and DNC is a big piece to the program for the day. @powersa has us cranking on DNC for a whole afternoon with a good number of people. That means we have to make sure this thing works well - even if it's not at 100% functionality.

@alukach @thebigspoon what do you think are the most important features to finish up by then? My immediate reaction:

* prepare the menu space to take in all of the turf functions - this would be a great day to have people adding those in! Blockers: #5, #40 
* establish a proper pattern for adding `<form>` elements for these turf operations. I'm not exactly sure how we go about doing this best, but it seems important to move forward. Exists in #2 already
* Style some basemaps? Choose a basemap? Thinking of things that someone without programming experience could work on and potentially choose good basemaps for the data.
* Clear documentation for install, #47 
* **SUPER** clear documentation on the architecture of the project (can add to wiki)
* Establish a proper workflow for people, including for ourselves to test forks properly and add tests
* what else?

I think @thebigspoon will be there to help people out, but I'm a little worried that we're not going to be in a place to have 15 (total guess) people working on this project (that's insane, right?). @powersa you mentioned this will likely split into multiple groups working on stuff. Any idea what you expect those to be?

alukach:
So what's the plan for the SpringFling Code Sprint?  What will people be working on?

One big thing that I'd like to get done before anything else (and have been slowly working towards), is having a much stronger set of testing. I've added tests for the `Menu` class in #55, but to me the big one is to get an actual set of functional tests for [the existing operations](https://github.com/cugos/drop-n-chop/blob/31d076ac586c6a93e09ab03b5d8bde3cac5641ae/src/js/controller/AppController.js#L21-L32). Basically, each of these operations should be provided actual input and ensure that the output is as to be expected.

I assume that adding operations will be the first thing people working on the project will want to do, and it is super trivial to add Turf operations right now. It will be import to have tests to a) ensure that nobody messed up existing functionality; and b) provide a pattern for people to follow to write tests for their new functionality.

I'm kind of operating on the mindset that no new functionality should be added until a base level of testing is completed and that underlying architecture is refined.  I think the architecture component is there after #53, but testing still has room.

svmatthews:
I agree, but also don't want to show up and be all, "hey we have to write tests for an application that doesn't do anything yet!" - seems *not very interesting* to me, personally.

[...]

PR #67, opened 2015-05-08
Menubar, menu, form, & operation refactor

svmatthews:
### What changed?

* including all changes from #63 
* **REFACTOR** how the `MenuBar`, `Menu`, and `Operations` work with each other. Now each is their own major `Class`, removing the need for nested `.addChild()`'s. The big change here is taking all of the turf operation parameters and putting them into `/operations/Geo.js`, which is essentially a reference object for all of that information that we use in `/forms/Forms.js` and `/operation/GeoExecute.js`. Each of these is now properly sending publications to the `AppController`.
* Menus are created with simple arrays, see [this line](https://github.com/cugos/drop-n-chop/compare/ops?expand=1#diff-dd6550659eb3f562e3daf4cb53c84280R24). That array creates new menu items, which when clicked grab the action id and grab the matching parameter in `Geo.js`. So... there are two places we have to reference the name now, but I think it makes the controller a lot cleaner and simpler to use.
* Commented `/spec/suites/Menu.js` since it technically no longer exists like this and was preventing me from actually doing this. I swear I'll learn how to fix it :stuck_out_tongue_closed_eyes: 
* sweet new DNC vector art

Here's the new js file tree

```
|-js
|---controller/
|-----AppController.js
|---dropzone/
|-----DropZone.js
|-----handlers/
|-------DropZone.FileReader.js
|---forms/
|-----Forms.js
|---layerlist/
|-----LayerList.js
|---mapview/
|-----MapView.js
|---menus/
|-----MenuBar.js
|-----Menu.js
|---notifications/
|-----Notifications.js
|---operations/
|-----Geo.js
|-----GeoExecute.js
```

### What's next?

* I think we are in a good position to move the "selection" into its own class object so we can reference selections from multiple locations (instead of relying on it being based on LayerList)
* Update tests
* add all of the turfs - and probably still think about #5 
* probably close #63


src/js/operations/Geo.js
```diff
@@ -0,0 +1,33 @@
...
+                default: 'miles'
+            }
+        ]
+    },
+
+    union: {
+        minFeatures: 2,
+        maxFeatures: 2,
+        description: 'Takes two polygons and returns a combined polygon. If the input polygons are not contiguous, this function returns a MultiPolygon feature.'
+    }
```
alukach: Good work on this @svmatthews, this seems a lot more sensible/testable.

svmatthews: Thanks @alukach! I'm going to make sure I've got tests down and ready before merging this - but hopefully ready before this coming Wednesday (#57) what do you think?

svmatthews:
@alukach I managed to get the tests working again! There are still some commented out sections (in the public functions portion) because this update no longer uses them. Probably worth giving a check.

I think this branch is ready to be merged - allowing us to start writing tests for all other pieces (ops, geo, etc.).

alukach:
@svmatthews nice! Will look through all of this when I get a chance.  I think this should make writing operation tests a lot more simple.

src/js/operations/Geo.js
```diff
@@ -0,0 +1,33 @@
+L.DNC.Geo = L.Class.extend({
```
alukach: Why the switch from `TurfOperations` to `Geo`? Do you predict that this project will always be a wrapper for `turf.js`? I had envisioned other operations not necessarily covered within `turf.js`, such as generating heatmaps.

This may be a premature concern, but I think it'd be wise to keep open the door for expanding beyond `turf.js`.

src/js/operations/Geo.js
alukach: Are we dropping `erase` for this refactor?

src/js/operations/Geo.js
svmatthews: That's funny that you say that, because that's my same feeling so I generalized the file name. I'm fine continuing it as Turf, but I figured at some point we'll have turf operations interacting with other geo operations so going one big Geo file was my approach.

We're both in agreement, just different file naming ideas :grin:

alukach: Ah, okay, I see where you're coming from.

Now that the execution is broken out from the operation, it seems like whether it's a `turf` operation or another type of operation is not as relevant.  Where it will be relevant is with the execution. I imagine that we'll sooner or later have to add an `executor` property to each `Geo` operation that points to different subclasses of `GeoExecute` or something.  

It's probably not worth focusing too much on this now, as we don't know which other geo-operations will be added in the future or how they'll need to be implemented.

src/js/controller/AppController.js
```diff
@@ -15,33 +15,43 @@ L.DNC.AppController = L.Class.extend({
...
-                }))
-            );
+        
+        this.menubar = new L.DNC.MenuBar( { id: 'menu-bar' } );
 
-        this.notification = new L.DNC.Notifications();
+        // new menu
+        this.menu = {
+            geo: new L.DNC.Menu('Geoprocessing', this.menubar, { 
+                items: ['buffer', 'union', 'erase', 'intersect']
```
alukach: If we're to merge this PR, we should probably either trim this list of the un-implemented `erase` and `intersect` or build out the tooling in `L.DNC.Geo` (I vote for the latter).

Interesting observation: if a menu is not defined, it will load the last options menu of the last operation clicked.

bug

svmatthews: Agreed! Let's remove and add as separate work.

PR #68, opened 2015-05-08
Change map ID and token

lyzidiamond:
- Changes map ID to `mapbox.streets`
- Changes access token


svmatthews:
Yay mapbox streets! Looks good to me.

PR #69, opened 2015-05-08
Update turf and mapbox.js library versions

lyzidiamond:
- Updates turf to `2.0.2`
- Updates Mapbox.js to `2.1.9`

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -2,12 +2,12 @@
 <head>
   <title>Drip. Drop. Drip.</title>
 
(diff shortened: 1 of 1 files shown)


powersa:
:+1:

svmatthews:
Good catch! :baseball:

svmatthews:
@lyzidiamond it looks like the CDN version you added isn't found for Turf. 

screen shot 2015-05-10 at 9 06 27 am

PR #70, opened 2015-05-08
Readme updates

lyzidiamond:
- Changes Leaflet.js to Mapbox.js in readme (you're using Mapbox.js)
- Grammar edits

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 **You drop. We chop.**
 
-Drop 'n Chop (DNC) is a browser-based GIS powered by [Leaflet.js](http://leafletjs.com) and [Turf.js](http://turfjs.org). DNC is currently a proof of concept and explores three hypothesis:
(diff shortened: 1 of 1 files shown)


svmatthews:
Truth, but we're relying heavily on the Leaflet architecture. What's the best way to specify? Is using Mapbox.js necessary?

lyzidiamond:
It's just the reality. Mapbox.js is a Leaflet.js plugin. By saying you're using Mapbox.js, it's implied that you're using Leaflet.js, because the two can't really be decoupled.

svmatthews:
:dancer: thanks for the clarification! It seems like it will be useful to continue with Mapbox.js since that gives us access to a suite of tools that will likely be used in the future.

lyzidiamond:
Thinking about it a little more... I think there's room for an **About** that explains how the application works w/r/t Leaflet's architecture. Let's talk about it more when we meet to chat about `CONTRIBUTING.md` @svmatthews

svmatthews:
:+1:

PR #71, opened 2015-05-08
move license to root level

svmatthews:
@alukach I'm moving the LICENSE to the root level. Was it intentionally in the `spec` directory?


alukach:
Hmmm... I have no idea how this ended up in `spec`. Good change

PR #72, opened 2015-05-08
Baselayer switcher

lyzidiamond:
- Adds a baselayer switcher to switch between:
  - Mapbox Streets (default)
  - Mapbox Outdoors
  - Mapbox Light
  - Mapbox Dark
  - Mapbox Satellite
- Adds control for baselayer switching to bottom right, not collapsed

The idea here is to be able to add custom layers with custom map IDs in the future, but this is just a start / proof of concept.


lyzidiamond:
I recognize that in the future if there was to be baselayer switching/adding customization it would need its own control outside of `MapView.js`, but like I said, this is quick-and-dirty.

svmatthews:
Boom! :boom: 

I think it's okay to not have feature ideas implemented at their fullest potential. Having these options up on the code allows us to start the conversation and think about how to implement effectively. Thanks for the reminder - sounds like a new issue can be created for "add your own base layer" :smile:

PR #79, opened 2015-05-12
Testing/operations

alukach:
What's changed:

- Small refactor of `L.DNC.GeoExecute`, pulling out logic that isn't directly related to executing an operation (ie putting results on the map) and placing that logic in `L.DNC.AppController`.  For the sake of testability, I think life would be easier if we work towards limiting the scope of classes and functions.  Additionally, having functions only rely on arguments as inputs (as opposed to global variables) and only using return values as output (rather than calling functions from the global namespace)

To be added:

- A test for the `buffer` operation


src/js/controller/AppController.js
```diff
@@ -41,20 +41,30 @@ L.DNC.AppController = L.Class.extend({
         this.dropzone.fileReader.on( 'fileparsed', this._handleParsedFile.bind( this ) );
         this.forms.on( 'submit', this._handleFormSubmit.bind(this) );
         this.menu.geo.on( 'click', this._handleGeoClick.bind(this) );
-        this.ops.geox.on( 'created', this._handleGeoResult.bind(this) );
```
alukach: `ops.geox` no longer sends back a `'created'` signal, it just [returns it's results directly](https://github.com/cugos/drop-n-chop/pull/79/files#diff-dd6550659eb3f562e3daf4cb53c84280R57).

src/js/operations/GeoExecute.js
```diff
@@ -54,25 +54,16 @@ L.DNC.GeoExecute = L.Class.extend({
 
         // if the new object is a feature collection and only has one layer,
         // remove it and just keep it as a feature
-        if ( newLayer.geometry.type == "FeatureCollection" &&
-             newLayer.geometry.features.length == 1 ) newLayer.geometry = this._unCollect( newLayer.geometry );
-
-        var mapLayer = L.mapbox.featureLayer(newLayer.geometry);
```
alukach: This type of logic is removed from this class.  `GeoExecute` isn't in charge of doing anything with the data, it just executes the operation and returns the results.  This should make testing a bit easier.

svmatthews:
:+1: from me on this one. Testing seems like it should be much easier this way. As you say "relying more on returns from functions would make it easier to treat each little chunk as a unit and to test on a more isolated level".

svmatthews:
Hmm.... looks like some tests didn't pass.

alukach:
@svmatthews there we go, forgot to use `.eql` for deep comparisons.

svmatthews:
Ah, good to know!

Issue #80, opened 2015-05-12
How do we handle turf functions that return something other than geometry?


svmatthews:
Currently all of our turf functions return geometry and we add as a new layer. How do we propose working with functions that are essentially mathematical functions that are more useful for attributes than new data layers?

Turf count is an example of a function that doesn't return new geometry. http://turfjs.org/static/docs/module-turf_count.html

svmatthews:
Also, how do we handle turf operations that don't require geometry and essentially let you build new geometry? Such as `square-grid` http://turfjs.org/static/docs/module-turf_square-grid.html

angele-smell:
Using turf.area, which is meant to return the area of 1 or more features in square meters, a layer is returned, but doesn't have geometry, and it doesn't look like the area data is even stored in that file. The example I tried was on a single polygon.

alukach:
#123 should open the door to handle non-spatial returns.  Someone will need to create a `notify` action (or something similarly named) on `L.DNC.AppController._handleResults`, returning the values to the user via a notification/popup/etc

PR #97, opened 2015-05-13
Refactor the way operations are bound to menu items

### LINKED ISSUE #64: Layers are starting to generate on layerList out of nowhere after operations complete
Look at the layers that come out of the union in this gif:

dnc-bufferz

`sf.geojson` and `east.geojson` are duplicated. Wtf!? Probably a result of some of the work on #63

alukach:
# What changed...

## Refactor operations
Prior to this code change, `L.DNC.AppController._handleFormSubmit()` assumed that all form submits were `L.DNC.Geo` operations (defined within `this.ops.geo`) to be executed by `L.DNC.GeoExecute` (defined within `this.ops.geox`).  Many future operations will not be geo operations (such as 'Save').  This is an effort to break that restriction by allowing the handlers of `clickedOperation` to be configured when the signal is attached.

To distinguish between handling results of geo-operations and non-geo-operations, a `createsLayer` property has been added to all geo-operations.  This should hopefully play nicely with #80.

## Add validation BEFORE forms appear
Operations are now validated before forms are rendered.  Errors will appear via notification.

## Bugfix: Prevent multiple layer being generated
A new listener to `submit` was being attached to a form every time a form was rendered. These listeners persist between forms, causing multiple layers to be generated on a single operation.  A hacky fix was added to clear listeners after the form is submitted.  This may fix #64. Not sure.

### DIFF
--- a/src/js/controller/AppController.js
+++ b/src/js/controller/AppController.js
@@ -58,6 +58,22 @@ L.DNC.AppController = L.Class.extend({
     */
     _handleOperationClick: function( opsConfig, e ) {
         var config = opsConfig.operations[e.action];
+        try {
+            opsConfig.executor.validate(
+                this.getLayerSelection(),
+                config
+            );
+        }
+        catch(err) {
+            this.notification.add({
+                text: err.message,
+                type: 'alert',
+                time: 2500
+            });
+            console.error(err);
+            return;
+        }
+
         var form = this.forms.render( e.action, config );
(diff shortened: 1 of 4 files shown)


alukach:
@svmatthews take a look at this one when you get a chance, some medium-sized changes in here.

alukach:
Note, upon further inspection, that change did not fix #64

Issue #88, opened 2015-05-12
TURF: Add envelope operation


svmatthews:
Follow [these instructions](https://github.com/cugos/drop-n-chop/wiki/Adding-a-Turf-function) for adding a new turf function.

Turf docs: http://turfjs.org/static/docs/module-turf_envelope.html

Issue #86, opened 2015-05-12
TURF: Add center operation


svmatthews:
Follow [these instructions](https://github.com/cugos/drop-n-chop/wiki/Adding-a-Turf-function) for adding a new turf function.

Turf docs: http://turfjs.org/static/docs/module-turf_center.html

cybersea:
When running center on two selected polygons, Drop-n-Chop outputs 3 points.  The naming convention of these outputs implies:
* center of the first poly
* center of the second poly
* center of both polys

But, the one that should be the center of both polys is actually the center of the last polygon selected.  

I think this function is supposed to give the geographic center of all features

My understanding from the turf docs, is that it is expecting a feature collection, not individual features.  So, this might be the problem.

PR #100, opened 2015-05-13
Added center, centroid and envelope.

Acab77:


alukach:
Just firing em off!  Nice.  Think you'd be up to the challenge of creating some tests for these?

alukach:
Relates to #86, #88

PR #104, opened 2015-05-13
Added zoom control to map

eldang:
@kempj2 requested a zoom control, and I also found myself wanting one, so I added it.

Caveat: by default leaflet's idea of topright and topleft clashes with the top nav bar, so I've added a bit of a kludgey fix by simply adding 50px of margin to the `leaflet-top` div.  I am receptive to suggestions for better ways to achieve this!


svmatthews:
Looks good to me! :eyeglasses:

Issue #81, opened 2015-05-12
TURF: Add bezier operation


svmatthews:
Follow [these instructions](https://github.com/cugos/drop-n-chop/wiki/Adding-a-Turf-function) for adding a new turf function.

Turf docs: http://turfjs.org/static/docs/module-turf_bezier.html

flott:
I didn't see this before, but I suppose I claimed it via my pull request!

PR #106, opened 2015-05-13
Add bezier function with optional parameters

flott:


alukach:
Looks good at first blush!  Any chance that you can write a test for this puppy? :dog:

alukach:
Relates to #81

PR #107, opened 2015-05-14
Avoid readding unchecked layers

alukach:
I believe that I've gotten to the bottom of #64.

When you add a file to the app, it creates a layer on the map (`L.DNC.app.layerlist._map._layers`) and adds the layer to the layerlist (`L.DNC.app.layerlist._layers`).  It seems that the problem is that when you uncheck a layer, it is removed from the map's list of layers but not the layerlist. When a new layer is created, `L.DNC.LayerList._update()` analyzes all layers within `this._layers` and adds those that are not on the map. This means that when a new layer is created, `update()` will re-add any layers within the layerlist that are unchecked.  This is why we see redundant layers being mysteriously added.

`update()` only seemed to be used within `addLayerToList()`.  When we know exactly which layer is being added, it seems unnecessary to look through all layers and add any layers that aren't on the map.  This PR effectively replaces `update()` with the two following lines:

```javascript
this._map.addLayer( layer );
this._addItem( obj );
````

This corrects the bug:
**before**:
bug

**after**:
bugfix

Unfortunately, there are a lot of tests written around `update()` so this PR will require some reworking of the tests.


svmatthews:
Thank the lord for you @alukach

svmatthews:
Is this PR current with your work in #108? Looks like the Travis errors are the same.

PR #109, opened 2015-05-14
adding grunt-sass, which is libsass and not ruby dependent, resolves #99

### LINKED ISSUE #99: Get rid of Ruby dependency.  Find a NodeJS Sass compiler.
Grunt errors when Ruby and Sass are not installed. This dependency is (could be) unnecessary.

svmatthews:
### What changed?

After some wicked dependency issues, it became clear that removing the ruby sass dependency would make the install process more simple for those computers without sass installed already. Now using `grunt-sass`, which uses libsass (C++) instead of `grunt-contrib-sass`, which uses ruby.

Thanks @thedeb for posting issue #99


eldang:
:clap:

svmatthews:
@alukach you'll want to do a new `npm install` for this one

Issue #117, opened 2015-05-15
Add Layer / Remove Layer buttons


alukach:
I think it'd be helpful to have an "add layer" and "remove layer" button.

**Add Layer(s)**
- When clicked, opens a file selection input box. Maybe I don't use a mouse? Maybe I don't like using drag/drop for file?
- This is where future options like "add layer from URL" would live. In this case, when clicked, would open a menu with "Add from computer", "Import from URL", etc.

**Remove Layer(s)**
- Pretty self explanatory, when pressed it removes all selected layers from layerlist, map, and memory.

I see these two buttons living at the bottom of the layerlist, as :heavy_plus_sign: and :heavy_minus_sign: symbols.  Not married to that UI, just the first that I thought of.  Hoping we can stay away from cramming too much into the menu-bar.

svmatthews:
Sounds good to me! Will have to essentially hook into the `fileReader` and create a `removeLayer` hook.

PR #123, opened 2015-05-23
Add/remove layer buttons

alukach:
## Description

As per #117, add some controls at the bottom of the layerlist to add and remove layers. The 'add' control is a menu, exposing operations to add a file in different ways.  The 'remove' control should be an operation itself, rather than a menu (I can only think of a single way that remove would work). This is a bit of a departure from the current menu capabilities, as a menu bar only allows the containment of menus, not operations.

bottom-menu

## What's Changed

- The direction in which a `menu-dropdown` appears in relation to a `menu` is now driven by the `menu` options.
- `div#dropzone` -> `div#sidebar`
- `.menu-dropdown` has `z-index` of `1`, to overlap `map`

## To do
@svmatthews curious what you think about the UI of the button and it's placement. Not super crazy about it and very open to hear suggestions about how to improve it.

- [x] Add tooling to allow remove button to be added to menubar (rather than adding menu to menubar)
- [x] Allow operation to not have intermediary form
- [x] Add upload file tooling
- [x] Add import file tooling
- [x] Wire remove button to remove item from layerlist / map


svmatthews:
Nice workin! As far as the UI goes, I'm not too concerned about it right now. We haven't really made any decisions as far as the design so far. I think eventually all of these will have to turn into icons with tooltips, but that's a ways down the road (maybe?).

How does upload from URL work?

alukach:
@svmatthews Does an ajax request to a provided URL and fetches the data.
uploadfromurl

It's in efforts of moving the project to allowing DC to act as a data viewer/editor.  I'd really like to get to a point where a user could access http://dropchop.io/?url=https://gist.githubusercontent.com/alukach/0bf4471e44a0243b66db/raw/39a3ad702cf08af1d5aeb606a098a0b4a4477033/union1.geojson, which would load the file automatically.  This would allow people to easily have "Edit this file with DropChop" links, if someone so chooses.

svmatthews:
:boom: !!! That's awesome.

Agreed on all of the above.

alukach:
Alright, I believe that this had reached feature complete.

completed

There were some notable changes to the operations, forms, and menu classes:

- Loading files via AJAX required that the operations be asynchronous. Execution class (`GeoExecute` and `FileExecute`) `execute()` methods now take in a callback as an arg and pass results to that callback, rather than returning the results.
- Rather than returning a single layer, operations pass (to the callback) a result package. This is an object that describes what the callback should do with a given feature. Currently, it only supports `add` and `remove`.  If you were to remove two layers and add one (such as if there were a merge operation), you would return `{add: [newLayerObj], remove: [oldLayerObj1, oldLayerObj2]}`. `_handleResults` then handles these layers accordingly.  I could imagine an `update` action being added, or a `notify` action to handle operations that have no geo-results (such as `count`). As such, operations no longer need the `createsLayer: true` property.
- Forms intake an `extra` param from an operation parameter config, which appends whatever string given into the generated input.
- Forms can be skipped if the operation parameter has the `disableForm: true` setting.
- You can give menubars custom classes (used to push menubar to bottom of layerlist)
- You can provide a menu a custom icon class via the `iconClassName` option.
- If a menu has no sub-items, it is treated as a button (such as the 'remove' button).

This still requires a few tests but is ready to test.  Will comment again when tests are complete.

mapsam:
:clap: @alukach!

alukach:
This guy still needs some testing, I'll make an issue for it.

mapsam:
Oh, sorry! I read "feature completion" as ... complete.

Issue #11, opened 2015-03-25
Allow a way to download/save a new file


yzimet:
Perhaps [FileSaver.js](https://github.com/eligrey/FileSaver.js) ([blog post](http://updates.html5rocks.com/2011/08/Saving-generated-files-on-the-client-side)) would be suitable
>an HTML5 saveAs() FileSaver implementation

svmatthews:
Thanks for this! Let me know if you get a chance to play around with this, otherwise I'll take a look soon.

morganherlocker:
A common way to do this is posting to the gist.github.com api with an anonymous gist, then changing the href on your "save button" to point to the response.

svmatthews:
This is what geojson.io does, yes? I definitely like how you can save in multiple formats with that service - are all of the formats pushed through gist or does it use different scripts for those different formats?

svmatthews:
O_O 

Just started reading [this file](https://github.com/mapbox/geojson.io/blob/0f70ff7262377aff2d5affadc7a2f2e0a978a374/src/ui/file_bar.js#L471-L479) and Mapbox has an `npm` module for all of the things. :tada:

svmatthews:
@aaronr how is this one coming along?

aaronr:
Started it... will finish it soon.... just trying to finish up some work work

svmatthews:
Oh definitely no rush - just didn't want to step on toes :mans_shoe:

aaronr:
Got basic geojson working well... and shapefile download kind of working.  Using the shpwrite utility but it is flaky (looks like known issues with it).  So... wrapped in a try catch just like they do with geojson.io for now.  I have had it work on like 1/4 of the json I have loaded into the tool... and very few of the turf produced geoms.  Will have to look closer.  If it fails it gives a message to the end user asking them to try another format (i.e. download in geojson as that works).
Each layer is downloaded in its own file.  Pretty straight forward.
I added a dependency directory under the js dir to hold these 3rd party libs.  I also have the lint util skip them as they puke when linting.  

https://github.com/cugos/drop-n-chop/commit/5bb4c2945e1e53289ce2457b717397eeef174113

This is on the filesave branch and is up to date with master as of me writing this so will easily merge.  Test it out or give me the OK and I can merge it in.

aaronr:
@svmatthews and @alukach ... thoughts on this?  I would love to merge this in at least as a starting point.

alukach:
[...]

PR #128, opened 2015-05-29
Starting to get file saving to work.  The shapefile saver is kind of …

mapsam:
…flaky from mapbox and is just as flaky on geojson.io... but it is a start.  We use a try catch statement for the export of the shapefiles... so if they fail at least we display a warning and move on.  Export to geojson works much more consistently.  In the geojson case we export each layer separately and increment the name of the file. This is for issue #11

@aaronr opening this to start working on tests


mapsam:
@alukach this is going to need some love with your refactor from #117

alukach:
Yeah, looks like we both made `FileExecute.js` (and some others).  Will do a merge when I get a chance.

aaronr:
I am happy to revisit this... but the longer we wait the more merge conflicts we will get.  I was still unclear if you wanted me to back out the inclusion of the web ready dependencies I added in favor of switching our whole system to Browserify.  Seems to me getting this in as is makes more sense and then add (or re-open) a ticket to go in and add Browserify support back in.  Then finally a ticket to move the existing download stuff to utilize browserify instead of the static includes.  Doing all of this at once seems a bit hard to keep a handle on.

alukach:
I guess this begets the question: should the Save tooling go down under the LayerList?  If not, should the Load File go up in the `File` menu?

@aaronr Agreed that Browserify refactor doesn't need to be a part of this PR

mapsam:
I regret pushing that commit ...

alukach:
Why the regret?  The merge looked okay to me.

mapsam:
@alukach @aaronr this is all working now - I've added a mock test file to fill out. Might need some help with this one. (like every test)

Issue #129, opened 2015-05-30
Shift + Select offset


alukach:
It appears that, when holding `Shift` between two selections of layerlist items, the outcome selection is offset by 1 (ie, selects the correct number of layer, but selects 1 above what it should and 1 short of what it should on the bottom).

@mapsam any ideas as to why this could be?

![offset-select](https://cloud.githubusercontent.com/assets/897290/7895570/737e558c-0653-11e5-872f-d4b607219748.gif)

mapsam:
Weird that this started happening - updated the calculations! 

![dropchop-selectfix](https://cloud.githubusercontent.com/assets/1943001/7929573/720d1784-08b4-11e5-95ff-2eaab98221c7.gif)

PR #131, opened 2015-06-02
fix selection bug, #129

mapsam:
This fixes the multi select bug #129. Hopefully. :stuck_out_tongue_closed_eyes:


alukach:
:+1: Looks good

mapsam:
Well, the selection will likely require some testing to be written for it ... I'll make it as a separate PR though. Let me know if you run into any issues with it.

One major issue is that the user cannot unselect everything right now. I'd like to see a standard screen click deselect the layers - but not sure if `window.on('click', function () { ... } );` is legitimate or not. What do you think?

PR #132, opened 2015-06-03
Incorporate browserify

### LINKED ISSUE #8: incorporate browserify?

mapsam:
This adds the `grunt-browserify` task to start building dependencies without having to download them and include them in our `src`. Unfortunately, not everything is on `npm` so we'll have to keep the `src/dep/` directory for now (currently only includes the `filesaver.js`) library. Closes #8 

This also adds `shp-write` and `turf` as dependencies in the `package.json` and we now build them into the minified application, instead of using a CDN. How do we feel about this? Only downside is that it slows down the build a bit.

This PR also includes changes in #128


aaronr:
Nice!

PR #134, opened 2015-06-10
If input checkbox of selected item is toggled, toggle all selected items

alukach:
selection-update

I think that this should improve upon the 'natural' feel of the layerlist interaction.

Going to need some testing love before merged.

### DIFF
--- a/src/js/layerlist/LayerList.js
+++ b/src/js/layerlist/LayerList.js
@@ -107,7 +107,7 @@ L.Dropchop.LayerList = L.Control.extend({
             }
             this._map.fitBounds(bounds);
         }
-
+        return obj;
     },
 
     /*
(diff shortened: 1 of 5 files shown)


src/js/layerlist/LayerList.js
```diff
@@ -176,10 +176,30 @@ L.Dropchop.LayerList = L.Control.extend({
     */
     _handleLayerChange: function(obj, e){
         var inputEl = e.target;
+        var selected = obj.domElement.getElementsByClassName('layer-name')[0].classList.contains('selected')
         if (inputEl.checked) {
-            this._map.addLayer(obj.layer);
+            // If changed element in selection, alter entire selection
+            if (selected) {
+                this.selection.list.map(function(obj){
```
alukach: I'm starting to think that basic `for` loops are better than `.map`.  Faster, possibly more intuitive to reader, no need for `.bind`.

mapsam:
Good idea!

alukach:
Fyi, still have some more tests coming down the pipe, let's hold off on merging.

alukach:
Alright, this is good to go!  Added some tests for the layerlist.

mapsam:
Looks great on my end!

PR #135, opened 2015-06-11
Develop against unminified bundle

alukach:
Some changes were made to the Gruntfile that resulted in `grunt` command minifying the code during development.  This makes debugging more difficult.

These changes ensure that minification doesn't take place until `grunt prod` is run (or `grunt deploy`).


mapsam:
Sorry about this - I'm still in my cowboy phase of writing code. Will get better!

PR #139, opened 2015-06-14
Add data from GET parameters

alukach:
This allows a user to load data files into Dropchop by appending data sources to the URL (either `gist` or `url`). A user can append one or many data sources (provided that the URL stays under ~2000 characters). A `url` can be inputted as either a URL encoded or a URL decoded string. A `gist` can be inputted as the URL to the gist (eg `https://gist.github.com/alukach/0bf4471e44a0243b66db`) or the gist id (eg `0bf4471e44a0243b66db`). A 'load from gist' option was added to the 'Add' menu.

load-from-url-get

gist-url

I'm pretty happy with how this turned out.  It will allow people to more easily use Dropchop as a data viewer (as mentioned in #123). It also makes it a lot easier to populate the webpage with data during development.

Needs some tests and then will be good to go.

### DIFF
--- a/spec/suites/L.Dropchop.AppController.js
+++ b/spec/suites/L.Dropchop.AppController.js
@@ -1,0 +1,89 @@
+describe("L.Dropchop.AppController", function () {
+
+    describe("initialize", function () {
+        var map_div;
+        var sidebar_div;
+
+        beforeEach(function () {
+            map_div = document.createElement('div');
+            map_div.id = "map";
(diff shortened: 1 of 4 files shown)


mapsam:
This is awesome. Nice work @alukach!

alukach:
Thanks for the merge, still have a few more tests to write but this shouldn't hurt.

mapsam:
Ah crap sorry I missed the `in-progress` label.

Issue #130, opened 2015-05-31
A 'complete workflow' design-lab.


alukach:
I think that DropChop has been coming together really well since first started. It's getting the bits and pieces needed to be usable (operations, downloading, etc). However, I'm unsure if there's a unified target at the moment.  This issue is to intended to spur discussion about what DC should become. A general design-lab. We can add features here and there, but I think it's helpful to have a complete user experience that we work towards.

When I was looking into how to add some raster operations into the project, I came across the [Turf: GIS for web maps](https://www.mapbox.com/blog/turf-gis-for-web-maps/) introductory post. What stood out the most out of the post was the last example:

![live-update](https://cloud.githubusercontent.com/assets/897290/7902859/696df556-0785-11e5-833b-4adf8ec59f63.gif)

To me, this is what the target of DropChop should be.  Everything with the above `gif` should be doable in DC.

## Workflow

The following is an imaginary workflow that hopefully someone would be able to follow using DC:

A user reaches [dropchop.io](http://dropchop.io). They then:

1. Upload a line representing the path of the Bay to Breakers race route. This produces layer `raceroute`.
1. Style the `raceroute` line to be bright pink.
1. Upload a file of Water Fountains within San Francisco. This produces layer `fountains`.
1. Style the `fountains` points with small black icons of a drop of water.
1. Run a `buffer` operation on the race route. This produces layer `raceroute_buffer`.
1. Run a `within` operation to find all drinking fountains within 500 feet of route. This produces layer `fountains_within_raceroute_buffer`.
1. Style the `fountains_within_raceroute_buffer` points to have larger blue icons of a drop of water.`

Okay, so that's pretty simple so far.  We're not far from that workflow. The only thing we're missing is applying style, which is somewhat addressed in issue #126.

Let's keep going.  The user then decides that they'd be willing to go 1000 feet off the route for a drinking fountain. They:

8. Alter the `buffer` operation's output to have a buffer output of 1000 feet.  The `within` operation is automatically rerun and `fountains_within_raceroute_buffer` is updated.

This is something that seems pretty cool and different than any other GIS. The idea that layers can be bound to eachother and updates cascade downstream. This could be implemented by having each layer subscribe to their parent layers. When a parent layer updates, it calls for all subscribing layers to rerun themselves.

Another one: The user remembers that they plan to make a quick detour to a coffee shop halfway through the race. They:

[...]

Issue #126, opened 2015-05-27
Color layers in layerlist


alukach:
A user should be able to control the color of a layer's corresponding geometry within the layerlist.

I'd prefer that they be able to access this control via a menu brought up from a right-click on a layer in the layerlist.  The color of the layer would represented in the layerlist (possibly as a 5px border left of the text). The color could also represent whether the state was visible or not (turn to grayscale or something...).

alukach:
Updating this to pertain to the entire style of the layer.  This could include putting the icon representing a point feature into the layerlist.

Additionally, representing visibility of the layer seems unnecessary, as the checkbox does that for us.

PR #148, opened 2015-06-17
pulling colors into variables

mapsam:
This PR adds a thematic color scheme to dropchop (not definite, but was good to work with) and pulls a lot of our shared variables into the `_variables.scss` file so we can update styles easily without having to skim through a ton of `.scss` files.

screen shot 2015-06-17 at 12 07 14 am


mapsam:
Aww man, why did this fail!?

alukach:
Looks good to me!

mapsam:
Woop! I also added a couple more fonts, a monospace and a new sans-serif (Roboto). Mostly, I was getting sick of staring at what we had :wink:

alukach:
One concern that I have with this color palette is that it may possibly clash-with/influence the visual styling of data that a user may want to add (assuming #126 and #130 get employed).  I think the data and base-layer itself should be the primary focus from a user's perspective (which is something that a really neutral palette supports well).

I do like the aesthetics of this, but if I'm processing and styling data that I intend to put on my orange-themed red-toned site, these colors could make it more difficult to mentally visualize what the styled data will look like when I put it on my website.

mapsam:
Good point. I'm going to push back a little on this one, just because doing this styling stuff is what's getting me excited right now. Since everything is broken into the `_variables.scss` file we can make changes quickly. We could even allow the user to switch between bold/light versions.

Issue #90, opened 2015-05-12
TURF: Add midpoint operation


mapsam:
Follow [these instructions](https://github.com/cugos/drop-n-chop/wiki/Adding-a-Turf-function) for adding a new turf function.

Turf docs: http://turfjs.org/static/docs/module-turf_midpoint.html

svmatthews:
I don't get it, can you please explain?

mapsam:
resolved in #161

Issue #89, opened 2015-05-12
TURF: Add extent operation


mapsam:
Follow [these instructions](https://github.com/cugos/drop-n-chop/wiki/Adding-a-Turf-function) for adding a new turf function.

Turf docs: http://turfjs.org/static/docs/module-turf_extent.html

mapsam:
Extent returns a bounding box, which can then be turned into a polygon using `bbox-polygon` function, but this is exactly [how `turf.envelop` works already](https://github.com/Turfjs/turf-envelope/blob/master/index.js#L58-L59) - so incorporating these two options would be redundant right now.

Issue #87, opened 2015-05-12
TURF: Add destination operation


mapsam:
Follow [these instructions](https://github.com/cugos/drop-n-chop/wiki/Adding-a-Turf-function) for adding a new turf function.

Turf docs: http://turfjs.org/static/docs/module-turf_destination.html

mapsam:
Resolved in #161

Issue #85, opened 2015-05-12
TURF: Add along operation


mapsam:
Follow [these instructions](https://github.com/cugos/drop-n-chop/wiki/Adding-a-Turf-function) for adding a new turf function.

Turf docs: http://turfjs.org/static/docs/module-turf_along.html

mapsam:
Resolved in #161

PR #149, opened 2015-06-17
adding turf functions

mapsam:
Going to clean up some issues here:

- [x] TURF: Add midpoint operation #90
- [ ] ~~TURF: Add extent operation #89~~
- [x] TURF: Add destination operation #87 
- [x] TURF: Add along operation #85 
- [x] TURF: Add explode operation #84 
- [x] TURF: Add simplify operation #82


mapsam:
Adding midpoint:

dropchop-midpoint

mapsam:
Destination:

dropchop-destination

mapsam:
`along`

dropchop-along

mapsam:
`explode`

dropchop-explode

mapsam:
Hmmm, stuck at the end with `simplify`. I've written tests and they pass, but when I execute the function in the app the feature is only replicated, and not simplified.

alukach:
@mapsam Took a look at `simplify`, seems like there were two problems:

1. We were using `parseInt` rather than `parseFloat` for any number input, meaning that any `tolerance` < `1` would be `0`. Effectively, disabling it.
1. Using a `select` for a boolean seems a bit weird. The value that comes from a `select` is always a string. We could do something like `if (value == 'false' || 'true') value = parse(value);`, but I still feel like a `checkbox` is a better representation of a boolean.

Before:
bad-simplify

After:
good-simplify

mapsam:
Nice debugging! Thanks for adding some extra love to this. Agreed on checkbox considering there are other functions that will have options like this.

Considering this ready to merge if it looks good to you.

alukach:
Done, good work on this!

PR #150, opened 2015-06-17
Logo

mapsam:
Just adding the logo to the top bar, includes an updated `copy` function in the Gruntfile for image assets within `src`


mapsam:
screen shot 2015-06-17 at 10 54 22 am

alukach:
One concern that I have with this one is that it increases the height of the topbar, taking up precious screen reality for the sake of branding. I suggest we limit the icon height to keep the topbar height to a minimum.

mapsam:
Good point! I wonder if we can look into ways of actually removing the top bar entirely? Let's close this and I'll do some UX thinking.

alukach:
Yeah, if we go full bore with the sidebar, then I can see the topbar only residing over the layerlist (similar to the way the "Add/Remove" menus are placed at the bottom.

Issue #142, opened 2015-06-15
Right-click menu on layers


alukach:
I think the ability to right-click on a layer and select operations would provide a more natural interface.  When right-clicking, they would view a menu of "layer operations".  Operations that come to mind:

* Rename
* Alter layer style (see #126)
* Alter layer properties (see #140)
* Zoom to layer
* Zoom to selection

I struggle to come up with a good rule of what operations should go in the top menu bar and which should go as a right-click menu. I'm thinking operations that alters the state of an existing layer should go in the right-click menu.  Anything that produces a new layer or performs analysis on a selection of layers should remain in the upper menu bar.  Basically, the menu bar is for things that run ON layers, right-click is for things that happen TO layers... if that makes any sense.

I could also see the Save operations going in the right-click menu (for some reason, to me it feels a bit out of place in the top bar), but I don't feel completely confident about that.

mapsam:
I agree - the right-click would be much more useful to show ON layers instead of having a mixture of operations on the upper menu.

I experimented with this for a bit and using [this](http://stackoverflow.com/questions/4909167/how-to-add-a-custom-right-click-menu-to-a-webpage) as an entry point

alukach:
Yeah, I skimmed through [the first link that came up from google](http://www.sitepoint.com/building-custom-right-click-context-menu-javascript/), it seems pretty do-able.  I like our menus and definitely think we could make some minor modifications to the `Menu` class to allow it to be a right-click menu.  Agreed that they should be defined in `AppController`

mapsam:
As we add more geo operations, I'm also starting to wonder if geo should take on it's own menu type. Potentially something that sits along the layerlist sidebar, rather than a main menu. Honestly if we can move away from menu-bar menus that would be awesome.

Here's what I'm thinking:

![screen shot 2015-06-17 at 12 07 14 am](https://cloud.githubusercontent.com/assets/1943001/8218752/f14d0382-14f9-11e5-922b-438d2785b694.png)

alukach:
Yeah, I think a side bar could be a good idea.

mapsam:
Going to make this another issue #151

mapsam:
Right click context menus were added in #211!

PR #153, opened 2015-06-20
worldCopyJump: true

alukach:
From the [docs](http://leafletjs.com/reference.html):

> With this option enabled, the map tracks when you pan to another "copy" of the world and seamlessly jumps to the original one so that all overlays like markers and vector layers are still visible.

Before:

panning-bad

After:

panning-good

I feel like this setting can feel a bit 'glitchy' (jumps around, that weird reload flash that occurs, etc) but that's a problem with Leaflet itself and not Dropchop.  Hopefully it should get better.


alukach:
I'm not 100% confident that this should be merged, would like to hear some feedback after someone else tests out this branch.

mapsam:
It definitely seems better this way than before - especially if people are losing data. This also warrants building a button that can zoom to layer extent if you have it selected - or at least all of the data extent if you don't have anything selected.

alukach:
Yeah, definitely into the idea of "zoom to selection". I imagine it would go in the right click menu (#142)

PR #154, opened 2015-06-21
Avoid global import of notifications

alukach:
The way we are importing from `L.Dropchop.app` feels weird (and makes it difficult to test).

I would prefer to solve this via signals but it seems like you have to connect the `notifications` listener to ever signal sender, which seems laborious.  Instead, I opted to pass the notification tooling to the executor as an option.  This pattern can be applied to anything that would call `notifications`.

### DIFF
--- a/src/js/controller/AppController.js
+++ b/src/js/controller/AppController.js
@@ -21,6 +21,8 @@ L.Dropchop.AppController = L.Class.extend({
         this.mapView = new L.Dropchop.MapView();
         this.dropzone = new L.Dropchop.DropZone( this.mapView._map, {} );
         this.layerlist = new L.Dropchop.LayerList( this.mapView._map, { layerContainerId: 'sidebar' } );
+        this.notification = new L.Dropchop.Notifications();
+
         this.menubar = new L.Dropchop.MenuBar(
             { id: 'menu-bar' }
         ).addTo( document.body );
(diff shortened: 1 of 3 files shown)


alukach:
@mapsam not feeling too confident about this design, open to hearing alternatives

mapsam:
What do you think about using notifications only for non-visual successes and failures? Since the map updates with new information and the layerlist adds a new layer, that might be enough confirmation for the user. This way we can reserve notifications for errors and other non-map related actions.

I'm mostly ambivalent to the design. I wonder if we could pass a `notification: true` parameter with the `.fire()` options to make it a bit more streamlined?

alukach:
I dunno, I like the verbosity that comes with the current notifications.  I'm pretty indifferent on that topic.

> I wonder if we could pass a `notification: true` parameter with the `.fire()` options to make it a bit more streamlined?

I'm not exactly sure what you mean about this, can you go into greater detail?  My concern was that it seemed like we were trying to get away with passing requirements components from Dropchop modules into other modules, but maybe it was the global imports that was the real pain? My ideal situation is where any class could fire some `"Notify"` signal that `L.Dropchop.app.notifications` would be subscribing to (it seems like the current set up is where you'd have to subscribe to each class' signals explicitly, which seems like a pain to maintain).  I'm not feeling so bad about it anymore, this seems to work and I think it's pretty readable/understandable on first glance.

mapsam:
I'm all for keeping it as-is!

> I like the verbosity that comes with the current notifications.

Agreed.

And yes, that would mean we'd pass information from one module to another, which is against our current architecture. I'm pretty ambivalent to this mostly, so might as well not restructure if it's not necessary?

alukach:
Okay, cool, marching forward.  I think I got myself in a mindset of trying to getting groundwork perfect before adding features but in reality we've already cleaned out most of the oddities from the codebase and can now focus more on adding features.

mapsam:
:+1:!

PR #156, opened 2015-06-24
Buildout tests for gist / url loading

alukach:
A series of tests to accompany #130.

Finally feeling strong about my JavaScript testing game :muscle:, I think this covers the features well.

Also, modified the format of the `resultsPkg` sent back from operations:

Old:
```
{
    add: [{
        name: "along_line.geojson",
        geometry: {"type":"Feature","geometry":{"type":"Point","coordinates":[-10.839132601858164,38.352329563649434]},"properties":{}}
    }]
}
```

New:
```
{
    add: [{
        name: "along_line.geojson",
        geojson: {"type":"Feature","geometry":{"type":"Point","coordinates":[-10.839132601858164,38.352329563649434]},"properties":{}}
    }]
}
```
It felt a bit confusing to use `geometry` as a key when the data has a `geometry` key within it.

### DIFF
--- a/spec/suites/L.Dropchop.File.js
+++ b/spec/suites/L.Dropchop.File.js
@@ -1,15 +1,232 @@
 describe("L.Dropchop.FileExecute Operations", function(){
-	var menu;
-	var ops;
-	var exampleData = true;
 
-	// TODO: test for geojson save
-	describe('Save as a geojson', function () {
-		expect(exampleData).to.eql(true);
-	});
+    var menu;
+    var ops;
+    var exampleData = true;
 
-	// TODO: test for shapefile save
-	describe('Save as a shapefile', function () {
-		expect(exampleData).to.eql(true);
-	});
+    describe('Save as a geojson', function () {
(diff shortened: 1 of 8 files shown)


spec/suites/L.Dropchop.File.js
```diff
@@ -1,15 +1,232 @@
...
+            xhr.restore();
+        });
+
+        it('requests provided url, calls callback', function () {
+            var callback = sinon.spy();
+            L.Dropchop.FileExecute.prototype.getRequest('http://google.com', callback);
+
+            expect(requests.length).to.equal(1);
+            expect(requests[0].url).to.equal('http://google.com');
+            // expect(callback.called).to.equal(true);  // TODO: We should assert that the callback is called with proper results data. This seems to fail for some reason.
```
alukach: This was the only part of the tests that I could get to operate properly.  Still not sure what's up with it.

alukach:
This one seems non-contentious, merging in.

mapsam:
:+1:

Issue #155, opened 2015-06-23
Need a favicon


alukach:
Dropchop is definitely needing a cleaver or something as a favicon.

mapsam:
Nice! Could also add an icon from cugos/geoglyphs

alukach:
Cool.  Geoglyphs is looking awesome so far.

mapsam:
Added a geoglyph favicon! 

<img width="221" alt="screen shot 2015-07-12 at 11 53 55 am" src="https://cloud.githubusercontent.com/assets/1943001/8639369/cfbf945e-288c-11e5-83ea-e510c31f94e5.png">

This required adding a new `assets` directory under `src` and creating a new `copy:assets` command to be run with Grunt. Right now these are only copied over when you run the command manually because I wasn't sure where to add this in the build process. Should it be part of the default `grunt` command somewhere?

PR #164, opened 2015-07-12
adding favicon from geoglyphs, just to start, relates to #155

mapsam:
Adds a favicon from Geoglyphs, working on #155


alukach:
Looks good!

PR #168, opened 2015-08-08
resolves #167 by only allowing one operation run per click

### LINKED ISSUE #167: Clicking on an operation icon numerous time queues multiple executions
Each time that an operation icon is clicked, the execution of the operation is staged. When the execution is finally triggered, it runs as many times as operations that were staged (if that makes sense).  See example:

operation-click-bug

This occurs between different operation types as well:

operation-click-bug2

mapsam:
* resolves #167 by checking for an operation execution in the `AppController.js` and only allowing it to be reset once an operation has been run.

@alukach this should do it, but I wouldn't be surprised if I'm missing something.

### DIFF
--- a/src/js/controller/AppController.js
+++ b/src/js/controller/AppController.js
@@ -91,6 +91,7 @@ L.Dropchop.AppController = L.Class.extend({
     ** Bind events to handlers
     **
     */
+    clicked: false,
     _addEventHandlers: function(){
         this.dropzone.fileReader.on( 'fileparsed', this._handleParsedFile.bind( this ) );
         this.fileOpsConfig.executor.on( 'uploadedfiles', this.dropzone.fileReader._handleFiles.bind( this.dropzone.fileReader ) );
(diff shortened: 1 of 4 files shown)


src/js/controller/AppController.js
```diff
@@ -129,7 +130,10 @@ L.Dropchop.AppController = L.Class.extend({
             return this._handleFormSubmit( opsConfig, e );
         } else {
             var form = this.forms.render( e.action, config );
-            form.on( 'submit', this._handleFormSubmit.bind( this, opsConfig ) );
+            if (!this.clicked) {
+                this.clicked = true;
+                form.on( 'submit', this._handleFormSubmit.bind( this, opsConfig ) );
+            }
```
alukach: I think that this introduces a new bug:

1.  `this.clicked` starts off as `false`
1. I click the `bezier` icon, `this.clicked` turns to `true`, `submit` button becomes bound to the `geoOpsConfig`
1. I click a non-`geoOpsConfig` operation (eg add file). At this point, the `execute` button is still bound to the `geoOpsConfig` despite the fact that that I want a `fileOpsConfig` operation.

new-bug

mapsam: Okay, that's what I figured it would do. The whole solution felt like a hack :wink:

PR #169, opened 2015-08-08
Put popups above map layer selection

alukach:
Before:
screen shot 2015-08-08 at 3 47 36 pm

After:
screen shot 2015-08-08 at 3 47 44 pm


mapsam:
Boomtown :boom:

Issue #144, opened 2015-06-15
Add 'my location' to map


alukach:
A user should be able to add their location as a layer to the map.

lyzidiamond:
Ooh, this is a neat idea. Relevant: https://github.com/domoritz/leaflet-locatecontrol

mapsam:
![dropchop-location](https://cloud.githubusercontent.com/assets/1943001/10437367/27ba0722-70f2-11e5-9f25-851c97f908a8.gif)

mapsam:
Muchas gracias @jonahadkins

alukach:
Awesome, I love this feature.  So many fun possibilities are going to come out of this layer.

mapsam:
![scheming](https://cloud.githubusercontent.com/assets/1943001/10438522/b3550ac8-70f9-11e5-8344-3c0485dfa13c.gif)

jonahadkins:
woah! it looks awesome! :raised_hands: :raised_hands: :raised_hands:

PR #171, opened 2015-09-13
Add locate me

jonahadkins:
this adds enhancement #144 for a locate me feature

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -17,6 +17,8 @@
   <link href="//fonts.googleapis.com/css?family=Open+Sans:400,700" rel="stylesheet" type="text/css">
   <link href='http://fonts.googleapis.com/css?family=Inconsolata:400,700' rel='stylesheet' type='text/css'>
   <link href="//maxcdn.bootstrapcdn.com/font-awesome/4.3.0/css/font-awesome.min.css" rel="stylesheet">
+  <link href='https://api.mapbox.com/mapbox.js/plugins/leaflet-locatecontrol/v0.43.0/css/font-awesome.min.css' rel='stylesheet' />
+  <link href='https://api.mapbox.com/mapbox.js/plugins/leaflet-locatecontrol/v0.43.0/L.Control.Locate.mapbox.css' rel='stylesheet' />
   <link href="//api.tiles.mapbox.com/mapbox.js/v2.1.9/mapbox.css" rel="stylesheet" />
   <link href="css/app.css" rel="stylesheet">
 
@@ -34,6 +36,7 @@
       L.Dropchop.app = new L.Dropchop.AppController();
     };
 </script>
-
+<script src='https://api.mapbox.com/mapbox.js/plugins/leaflet-locatecontrol/v0.43.0/L.Control.Locate.min.js'>
+  </script>
 </body>
 </html>
--- a/src/index.html
+++ b/src/index.html
@@ -25,7 +25,7 @@
 
 <div id="sidebar"></div>
 <div id="map"></div>
-<a href='#' id='geolocate' class='ui-button'>Map My Location</a>
+<a href='#' id='geolocate' class='ui-button'>MAP MY LOCATION</a>
 
 <!-- build:[src]:dev js/bundle.js -->
 <script src="js/dropchop.min.js" type="text/javascript"></script>
--- a/src/index.html
+++ b/src/index.html
@@ -17,8 +17,6 @@
   <link href="//fonts.googleapis.com/css?family=Open+Sans:400,700" rel="stylesheet" type="text/css">
   <link href='http://fonts.googleapis.com/css?family=Inconsolata:400,700' rel='stylesheet' type='text/css'>
   <link href="//maxcdn.bootstrapcdn.com/font-awesome/4.3.0/css/font-awesome.min.css" rel="stylesheet">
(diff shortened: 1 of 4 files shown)


jonahadkins:
my first time installing node, grunt, etc  - contribution instructions were great! :100: 

locate feature references locate cdn with top left location - will need to add to css file for custom location.

mapsam:
Sweeeeeeet! Thanks for doing this :dancer: I'm working on a big overhaul in the `alpha` branch right now, which might mean the `leaflet-locatecontrol` will be installed differently at some point (via `npm` instead of CDN), but hot diggity glad you were able to add it! 

Not sure why the checks are failing here, though.

alukach:
Will this add the location as a layer or only zoom to a persons location?  It looks like just the latter? I think both are valuable, so thanks for putting this in!

mapsam:
Ohhh location as a layer would be awesome.

jonahadkins:
Just zooms to the persons location :disappointed: - the users location can probably be obtained from the marker or centroid extent, but that's beyond my skills (for now). 

i wasn't sure where to add the library to the project - there are a few css properties we can use to move the locate me button underneath the zoom in/out control.

mapsam:
No sad faces! This is all :smile:s. I've noted "turn user location into layer" as a bonus feature in the alpha and it can be unrelated to this PR.

As for adding libraries, we've gone back and forth between using CDNs, `npm` & bundle, and `require()` in our files. I'm leaning towards `npm` & bundle because it lands in the middle of efficient and manual. [Here's what it looks like](https://github.com/cugos/dropchop/blob/alpha/gulpfile.js#L34-L40) in the `alpha` branch right now after you `npm install XXXXXXX --save` and bundle it.

jonahadkins:
ill wire it up to use npm & bundle :+1:

mapsam:
Cool!

mapsam:
Hey @jonahadkins it looks like there is a relatively quick way to do this using the `locate()` function in `mapbox.js`. https://www.mapbox.com/mapbox.js/example/v1.0.0/geolocation/ - might not need to introduce another library! This also returns geometry, which could be used subsequently to make a new layer.

jonahadkins:
Even better - Ill work on that then

jonahadkins:
currently having beginner issues with [this._map.locate();](https://github.com/jonahadkins/dropchop/blob/add_location_%23144/src/js/mapview/MapView.js#L54)  throwing a `Uncaught TypeError: Cannot read property 'locate' of undefined` error

npeihl:
@jonahadkins I wonder if the `this` variable isn't available in the function due to closure issues? Unfortunately, I'm not sure what the best way to handle `this` is. Maybe @mapsam or @alukach have ideas.

alukach:
@jonahadkins Haven't had a time to actually checkout your code, but it sounds likes you `this` object isn't what you think it is in that function.  It's an awkward JS issue that is can bit strange if you haven't experienced it before. The error you posted is basically saying that `this._map` is `undefined`.  Basically, `this` refers to the scope within that `function (e)` that you're defining, so it won't be the same as the `this` on the lines above.  There are two common ways to fix this:

* Create a point of `this` outside of the function that won't be overwritten when you enter the function. ie:
```javascript
if (!navigator.geolocation) {
    geolocate.innerHTML = 'Geolocation is not available';
} else {
    var _this = this;  // Assigning the old 'this' to a variable that won't be overwritten
    geolocate.onclick = function (e) {
        e.preventDefault();
        e.stopPropagation();
        _this._map.locate();  // Explicitely using the old 'this'
    };
}
```
* Use [`bind`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/bind) to tell the function to use the class' `this`:
```javascript
if (!navigator.geolocation) {
    geolocate.innerHTML = 'Geolocation is not available';
} else {
    geolocate.onclick = function (e) {
        e.preventDefault();
        e.stopPropagation();
        this._map.locate();
    };
}.bind(this)
```

I'm no expert, but I feel like either of those should help.  I think option 2 is a bit cleaner.

Ps. Watch out for the indentation in that class. It's functionally the same, but a well indented chunk of code is much easier to read.

mapsam:
Yes, thanks @alukach! The confusion of `this` is mostly a relic in the `alpha` branch, which makes jumping into the code a bit more straightforward. 

It will look more like this in `alpha`, with `dc` being short for `dropchop`:

```JS
var myLocation = dc.map.m.locate() // can use this anywhere, without referring to "this"
```

jonahadkins:
omg y'all *IT WORKS*

i get a warning from chrome dev tools:

`getCurrentPosition() and watchPosition() are deprecated on insecure origins, and support will be removed in the future. You should consider switching your application to a secure origin, such as HTTPS. See https://goo.gl/rStTGz for more details.`

mapsam:
*aw paying money for https:// sounds like a fun way to spend my day*

alukach:
So, allegedly you can set up SSL w/ custom domains for free: https://sheharyar.me/blog/free-ssl-for-github-pages-with-custom-domains/.  Will add an issue.

mapsam:
Whoa. Nice!

jonahadkins:
FYI - Everything Works - I think the button might need some JS help to stay put (the previous library we used did) 
What else can I do / need to do on this branch?

PR #173, opened 2015-09-15
query the overpass api

mapsam:
**this merges into the `alpha` branch**

currently only querying nodes, but we should get ways and other things!

dropchop-overpassnodes

to consider complete:

- [x] query for `way`s, not just `node`s
- [x] handle errors & other responses
- [x] set a name for the layer as a query parameter


alukach:
Okay, this is crazy awesome. One of those things that makes me stop and ask "wait, technology can do that now?"

mapsam:
I know! I mean, the overpass API is nothing elegant. I still have next to no idea what I'm doing when building the queries for it:

```
http://overpass-api.de/api/interpreter?[out:json];node(57.7,11.9,57.8,12.0)[amenity=bar];out;
```

????? :stuck_out_tongue_closed_eyes:

mapsam:
Inspiration came from this: https://github.com/perliedman/query-overpass

PR #191, opened 2015-10-15
Add function to load features from ArcGIS feature service

npeihl:
**Ready to merge**

- [x] Rename `toGeoJSON` function [here](https://github.com/npeihl/dropchop/blob/arcgis-feature-services/src/js/ops.file.dropchop.js#L142) from `esri2geo` to something more descriptive
- [x] Add tests
- [x] Define `outfields` parameter in feature service query as either `'*'` or as text field option
- [x] Option to load only features in current map view
- [x] Add support for jsonp fallback where CORS is not enabled on server

Other suggestions?

### DIFF
--- a/src/js/ops.dropchop.js
+++ b/src/js/ops.dropchop.js
@@ -38,6 +38,7 @@ var dropchop = (function(dc) {
           'upload',
           'load-url',
           'load-gist',
+          'load-arcgis',
           'load-overpass',
           'location'
         ]
@@ -60,7 +61,7 @@ var dropchop = (function(dc) {
           'rename',
           'remove',
         ]
-      },      
+      },
       'info'
     ];
 
--- a/src/js/ops.dropchop.js
+++ b/src/js/ops.dropchop.js
@@ -74,9 +74,9 @@ var dropchop = (function(dc) {
       var action = setup[i];
       if (typeof setup[i] !== 'object') {
         if(dc.ops.file[action].type === 'break') {
-        var $breakSpace = $('<div>').addClass('menu-action-break');
-        leftMenu.append($breakSpace);
-      } else {
+          var $breakSpace = $('<div>').addClass('menu-action-break');
+          leftMenu.append($breakSpace);
+        } else {
           leftMenu.append(buildMenuButton(action));
         }
 
--- a/src/js/ops.dropchop.js
+++ b/src/js/ops.dropchop.js
@@ -1,5 +1,5 @@
 var dropchop = (function(dc) {
-  
+
   'use strict';
 
   dc = dc || {};
@@ -29,6 +29,7 @@ var dropchop = (function(dc) {
     $(dc).on('operation:file:load-gist', dc.ops.file['load-gist'].get);
     $(dc).on('operation:file:load-url', dc.ops.file['load-url'].get);
     $(dc).on('operation:file:load-overpass', dc.ops.file['load-overpass'].get);
+    $(dc).on('operation:file:load-arcgis', dc.ops.file['load-arcgis'].get);
     $(dc).on('operation:file:rename', dc.ops.file.rename.callback);
 
     // setup ops file
@@ -118,7 +119,7 @@ var dropchop = (function(dc) {
 
     // create name array
     data.name = dc.util.concat(nameArray, '_', operation);
-    
+
     return data;
   };
 
@@ -158,7 +159,7 @@ var dropchop = (function(dc) {
       btn.removeClass('operation-inactive');
       btn.prop('disabled', false);
     }
-    
+
   };
 
   return dc;
--- a/src/js/util.dropchop.js
+++ b/src/js/util.dropchop.js
@@ -1,11 +1,11 @@
 var dropchop = (function(dc) {
-  
+
   'use strict';
 
   dc = dc || {};
-  
+
   dc.util = {};
-  
+
   dc.util.removeFileExtension = function(string) {
     string = string.replace(/\.[^/.]+$/, "");
     return string;
@@ -49,7 +49,7 @@ var dropchop = (function(dc) {
         console.log(reader);
         shp(reader.result).then(forwardToGeoJsonIo);
       };
-    } else { 
+    } else {
       reader.readAsText(file, 'UTF-8');
       reader.onload = function() {
         $(dc).trigger('file:added', [file.name, JSON.parse(reader.result)]);
@@ -107,6 +107,14 @@ var dropchop = (function(dc) {
     return sw.lat+','+sw.lng+','+ne.lat+','+ne.lng;
   };
 
+  dc.util.getEsriBBox = function() {
+      // grrrr...
+      var bounds = dc.map.m.getBounds(),
+          sw = bounds.getSouthWest(),
+          ne = bounds.getNorthEast();
+      return sw.lng+','+sw.lat+','+ne.lng+','+ne.lat;
+  };
+
   dc.util.uncollect = function(fc) {
     return fc.features[0];
   };
@@ -131,7 +139,7 @@ var dropchop = (function(dc) {
     }
 
     return newFC;
-    
+
   };
 
   dc.util.loader = function(yes) {
--- a/src/js/util.dropchop.js
+++ b/src/js/util.dropchop.js
@@ -39,6 +39,20 @@ var dropchop = (function(dc) {
     xhr.send();
   };
 
+  dc.util.get = function(url, data, cors) {
+      var params = {
+          data: data,
+          dataType: 'json'
+      };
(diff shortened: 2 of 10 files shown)


mapsam:
I see the cross origin support task @npeihl - was the `dc.util.xhr` loader not working for a particular URL?

npeihl:
@mapsam I'm still unsure how to troubleshoot CORS issues. The URL I tried to use was [this](http://sjcgis.org/arcgis/rest/services/Shoreline/Green_Shores/MapServer/2). It may be that CORS isn't enabled on my server or can I do something about it on the client end? 

A client solution would be preferable to avoid similar issues with other organizations who aren't amenable to making changes to their servers.

mapsam:
Hmmm, maybe we can call on @ngoldman for some insider tips? What's the best way to start querying an Arc feature service without running in to CORS issues?

jgravois:
CORS functioning correctly requires that *both* the browser and server support it.

this browser based [test](http://client.cors-api.appspot.com/client) is a little unintuitive, but you can see that @npeihl's url doesn't trigger the same [200](http://client.cors-api.appspot.com/client) message as http://sampleserver6.arcgisonline.com (which supports CORS).

in esri leaflet, we assume support for CORS, but we also expose an option to fallback to JSONP for servers/browsers that don't support it.

in your situation, you could do something similar (either expect the end user to let you know, or run a quick test yourself).

ngoldman:
@jgravois to the rescue :+1:

npeihl:
Thanks @jgravois. That makes sense now. I administer the server for the aforementioned URL, so I can work on enabling CORS there. 

It seems King County has already enabled CORS as [this url](http://gismaps.kingcounty.gov/arcgis/rest/services/Parks/KingCo_Park_Finder/MapServer/2) works fine. 

I wonder how common it is that organizations don't enable CORS? Should we also have JSONP as a backup?

mapsam:
Thanks @jgravois! Sounds like a bit of this is out of our hands re: CORS-disabled URLs - building a JSONP fallback sounds like a good idea to me, rather than requiring the user to let us know.

ngoldman:
jsonp++

jgravois:
@mapsam anytime.

> ha! of course ESRI's bounding box needs to be in a different order ...

we prefer XY in map constructors too :smile:

mapsam:
@jgravois oh man ... I phrased that poorly. Being a grump about [this stuff](http://www.macwright.org/lonlat/) :stuck_out_tongue_closed_eyes:

jgravois:
no worries at all!  we've got thicker skin than that. :palm_tree:

mapsam:
Looking good @npeihl!

> Fix whitespace in diffs due to different editor preferences

Sounds like we should establish a strong linting regiment in gulp so errors are brought up before a push exists. I'll open a separate issue, don't worry about it for yours!

npeihl:
I'll try to work on this more later this week. I'm crunched on time trying to ready my presentation to CUGOS on Wed. :scream_cat:

mapsam:
Ahhh @powersa stealing my contributors away! :wink:

mapsam:
How's this coming along @npeihl? Need any help?

npeihl:
I've gotten everything done and working except tests. I can try to complete tests this evening. It would be nice to have a second pair of eyes do a code review and check.

mapsam:
Definitely. I'll take a look this evening and am happy to help out with tests!

mapsam:
Just got this running on my machine - looking super great @npeihl! Well done. I'm going to play with some tests and will keep you posted.

npeihl:
Ugh, OK, I'll admit defeat. I can't figure out the best way to write a test for this. I tried writing a test for the `callback` function, but it doesn't return anything. So then I tried writing a test for the `get` function, but that gets more complicate trying to mock an `XMLHttpRequest`. 

Maybe it's not even totally necessary to have a test given how the function basically just wraps other library functions (`$.ajax` and `esri2geo`). Any help or advice would be appreciated. 

Sorry this PR has taken this long. Things keep getting in the way.

mapsam:
Definitely not a problem. The way I got around writing a test for xhr was by *not writing a test* :wink: 

Since we are using `$.ajax()` I'm not terribly concerned if that function itself works. Why don't we start by unit testing what we can (i.e. with a fake, expected response from Arc, converting that, and adding to the map). I'm happy to get tests written, unless you're itching for it.

npeihl:
@mapsam I had given some thought into passing a fake ArcJson response into the `callback` function. [This](https://gist.github.com/npeihl/9d920e740560be01d7aa) is as far as I got. But since `callback` doesn't return anything, I'm stuck trying to establish an assertion. I don't think I can unit test the `get` function without some sort of mock server.

mapsam:
That's on the right track. What I've learned from a lot of @alukach's tests is that we can play around with good/bad versions of an ArcJson response to ensure we are catching everything. Might be worth making some of the functionality within `dc.ops.file['load-arcgis']` accessible for the tests. For example, testing for `jsonp`:

```javascript
dc.ops.file['load-arcgis'].checkJsonp = function(res) {
  // assumes an arcJson response
  if(res[2] === 'JSONP') {
    return false;
  } else {
    return true;
  }
};
```

then you can access this like

```javascript
var cors = dc.ops.file['load-arcgis'].checkJsonp(parameters);

// OR maybe `this` will work within the function?

var cors = this.checkJsonp(parameters);
```

What do you think about that?

npeihl:
@mapsam That makes sense. I'll take a stab at this tomorrow evening again.

mapsam:
:+1:

npeihl:
If codeship ci passes, I'd love to get this merged. Please let me know if you have any concerns on the unit tests.

src/js/ops.dropchop.js
```diff
@@ -72,9 +74,9 @@ var dropchop = (function(dc) {
       var action = setup[i];
       if (typeof setup[i] !== 'object') {
         if(dc.ops.file[action].type === 'break') {
-          var $breakSpace = $('<div>').addClass('menu-action-break');
-          leftMenu.append($breakSpace);
-        } else {
+        var $breakSpace = $('<div>').addClass('menu-action-break');
+        leftMenu.append($breakSpace);
+      } else {
```
alukach: This indention seems a bit off. Suggest indenting 2-spaces more.

alukach:
@npeihl looks good on my initial read-over.  Haven't had a chance to test it in action (nor do I have access to a ArcGIS feature server) but I love the idea of this feature!  Nice work!  I wouldn't let codeship stop you, it seems to be really inconsistent right now, I'm not really sure what's up (looks like this build is eternally in 'waiting' status).  I've added a task for odd failures that I've been receiving: #226.

mapsam:
Looks great! :+1:

jgravois:
@alukach you can find lots of fair game public sample feature services in our esri leaflet [examples](http://esri.github.io/esri-leaflet/examples/simple-feature-layer.html).

src/js/ops.dropchop.js
npeihl: I've fixed this in my latest commit.

npeihl:
OK. I'm happy with this and would like to see it merged. I don't have merge access so I'd be grateful if someone could for me. :smile_cat:

alukach:
@npeihl Nice work!  This should be fun.

@jgravois Thanks, that was helpful!

mapsam:
:clap: well done!

src/js/util.dropchop.js
```diff
@@ -145,17 +159,28 @@ var dropchop = (function(dc) {
...
+    return sw.lng+','+sw.lat+','+ne.lng+','+ne.lat;
+  };
+
   dc.util.uncollect = function(fc) {
     return fc.features[0];
   };
 
+  dc.util.esri2geo = function(arcjson) {
+    return toGeoJSON(arcjson);
+  };
```
alukach: @npeihl Hey, I have a failing test:

```bash
FAILED TESTS:
  util.dropchop.js
    ✖ dc.util.esri2geo()
      PhantomJS 1.9.8 (Mac OS X 0.0.0)
    ReferenceError: Can't find variable: toGeoJSON
        at /Users/anthony/Projects/map-experiments/drop-n-chop/dist/static/js/dropchop.js:2
        at /Users/anthony/Projects/map-experiments/drop-n-chop/test/spec/util.spec.js:75
```

I think it comes down to this line, where does `toGeoJSON` come from?

npeihl: @alukach I should be part of the `esri2geo` package seen [here](https://github.com/cugos/dropchop/blob/master/package.json#L19). Maybe run `npm install` again?

alukach: @npeihl ah, great, that did the trick, thanks!

PR #192, opened 2015-10-16
Add support for dropchopping a zipped shapefile

### LINKED ISSUE #10: turn shapefiles into geojson in browser
using https://github.com/calvinmetcalf/shapefile-js

jczaplew:
Sorry about the weird diff...apparently Atom likes to add a line break to the end of files and mess around with whitespace.

Added the ability to drag and drop a zipped shapefile using [shapefile-js](https://github.com/calvinmetcalf/shapefile-js). Fortunately it was already a dependency and mostly coded, just had to wire a couple of things together.

The shapefile-js function previously used `forwardToGeoJsonIo` as a callback, which I modified to instead trigger a `file:added` event. Not sure if that breaks anything or not, but it didn't seem like it.

Also had to tweak the gulpfile to make sure that `shapefile-js` is compressed into the dependencies.

Hopefully closes #10. Let me know what you think!

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -37,6 +37,7 @@ var vendorJS = [
   './node_modules/jquery/dist/jquery.js',
   './node_modules/browser-filesaver/FileSaver.js',
   './node_modules/shp-write/shpwrite.js',
+  './node_modules/shpjs/dist/shp.min.js',
   './src/lib/shp-2-geojson.js',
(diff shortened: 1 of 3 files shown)


mapsam:
Excellent work! Just pulled it down and it works great (and really fast).

The `forwardToGeoJsonIo` was leftover from toying around with some other examples, going with the `file:added` trigger is exactly right.

:clap: for putting this together!

Issue #194, opened 2015-10-17
State preservation with URL-loaded layers


jczaplew:
If the layer is loaded from a URL that contains an ampersand, it is not parsed correctly.

For example `?url=https://macrostrat.org/api/v2/paleogeography?age=250&format=geojson_bare` will request `https://macrostrat.org/api/v2/paleogeography?age=250` because the search is split on the ampersand.

mapsam:
@jczaplew it might be worth writing an updated test for this, but definitely fixed in #195!

PR #195, opened 2015-10-17
Rework URL parser to handle URLs that contain ampersands

jczaplew:
This is a fix for #194 that deals with preserving layers from URLs that contain ampersands. Instead of splitting on `&`, a regex split is used to check for `gist=` and `url=`. The major downside to this method is that the regex will need to be updated if other data types are to be preserved using the URL. 

However, it retains original functionality to support things like `?gist=12345&gist=67890&url=https://foo.com/something=mug&mug=something`

### DIFF
--- a/src/js/util.dropchop.js
+++ b/src/js/util.dropchop.js
@@ -66,15 +66,26 @@ var dropchop = (function(dc) {
   };
 
   dc.util.jsonFromUrl = function() {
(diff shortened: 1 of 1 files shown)


mapsam:
Good fix! Thanks @jczaplew 

Btw, going to add you as a CUGOS *chopper* so you can stop forking and pulling.

PR #204, opened 2015-10-21
Make sure URI updates 

jczaplew:
This ensures that the URI updates when layers are added and removed from a gist or URL. 

The main change is that when a layer is added from a gist or URL, the type and URL are passed to the `file:added` event, which then stores those attributes of the layer in `dc.layers.list`. When a layer added or removed, a new function `dc.util.updateSearch` is called which sorts through `dc.layers.list` to find layers that have a `type` and `url`, assembles a new `search`, and pushes it to the URL if it is different from the current one. 

I'm unsure of the best way to write tests for this functionality - anyone have any suggestions?

Also, no idea why my code editor automatically tries to remove whitespace...it makes for these really messy diffs...really sorry about that.

If someone could look this over and test it out that would be great!

### DIFF
--- a/src/js/layerlist.dropchop.js
+++ b/src/js/layerlist.dropchop.js
@@ -1,5 +1,5 @@
 var dropchop = (function(dc) {
-  
+
   'use strict';
 
   dc = dc || {};
@@ -44,7 +44,7 @@ var dropchop = (function(dc) {
         $(this).removeClass('layer-new');
       });
 
-    
+
     var layerType = $('<span>').addClass('layer-type-image sprite sprite-layer-'+layerTypeIcon(layer.type));
     var checkbox = $('<input>').addClass('layer-toggle').prop({'type': 'checkbox', 'checked': true});
     var remove = $('<button>').addClass('layer-action layer-remove').html('<i class="fa fa-times"></i>');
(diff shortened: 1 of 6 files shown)


mapsam:
Awesome! I think the best way to test this is by unit testing the `dc.util.updateSearch` function on it's own. I'll give it a crack.

mapsam:
The only thing I'm noticing this changes is the look of the layer images when changing `layer.type` to something that doesn't match the [switch statement cases in the layerlist](https://github.com/cugos/dropchop/pull/204/files#diff-326407055281e753efbfebd19fbf7d4cR154) so you get layers that have an incorrect image representation. Here's an image:

screen shot 2015-10-20 at 8 05 43 pm

You can see all of the layers are given the default sprite value because they come in as `gist` instead of `FeatureCollection` or `Polygon`. I'm not terribly convinced that these images are great, considering a feature collection could be anything and it's tough to guess what the objects are without fully looping through them. What do you think? Worth keeping? We could add to the sprite for `gist` or `url` if that would help? I kind of like that idea.

mapsam:
Otherwise this is working great! Now to the tests ...

mapsam:
Test written! Seems to be working nicely...

```javascript
it('dc.util.updateSearch()', function() {
  var url = '/?gist=333';
  window.history.pushState(null, null, url);

  expect(window.location.search).to.equal('?gist=333');
  
  // adding a new layer should update, and remove the previous since
  // it doesn't actually exist as a layer
  dc.layers.add({}, 'new gist', gj, 'gist', '444');
  expect(window.location.search).to.equal('?gist=444');

  dc.layers.add({}, 'from a url', gj, 'url', 'http://google.com');
  expect(window.location.search).to.equal('?gist=444&url=http://google.com');
});
```

mapsam:
I've added this update to the CHANGELOG: https://github.com/cugos/dropchop/wiki/CHANGELOG

mapsam:
@jczaplew feel free to merge this in since the test is running :ok_hand:

jczaplew:
Added another commit that fixes the conflict with `layer.type` that was preventing the correct icon from being used in the layer panel. 

Thanks for writing tests! I really appreciate it - I'm a newb at testing ;-)

mapsam:
:claps:

PR #208, opened 2015-10-24
Nested menus

### LINKED ISSUE #197: turn "load data" buttons into a single option
With the incorporation of @npeihl's Arc feature service we are going to have at least 5 ways to add data to the app: `local files`, `url`, `gist`, `overpass`, and `feature service`. That's getting to be a lot of options on the left-hand bar, and the icons aren't going to be very versatile as we get into more company-specific options.

I propose turning all of the buttons into a "load data" button that has specific ways to load data that can be selected in the popup (or something). Might require developing a new menu type, or further developing the `form.dropchop.js` file to allow for form responsiveness based on what the user is selecting, but I think it will clean up the sidebar quite a bit.

mapsam:
This builds in the ability to make the left menu space have collapsable menus in order to preserve real-estate as we create more and more options.

A relatively major difference is that the menu is now built from a `setup` object in `dc.ops.setup` instead of looping through the `dc.ops.file.OPERATION` objects. This menu setup, located [here](https://github.com/cugos/dropchop/compare/nested-menus?expand=1#diff-1454db8304b25cb3d6b99183c0f18697R32) gives us the option to either throw in direct operations (as strings) or objects with actions, which turn into nested menus when you hover over them. Here's what it looks like:

dropchop-collapseablemenu

This should fix #197 for the time being.

### DIFF
--- a/src/js/ops.dropchop.js
+++ b/src/js/ops.dropchop.js
@@ -29,9 +29,9 @@ var dropchop = (function(dc) {
     $(dc).on('operation:file:load-overpass', dc.ops.file['load-overpass'].get);
     $(dc).on('operation:file:rename', dc.ops.file.rename.callback);
 
-    var leftMenuSetup = [
+    dc.ops.setup = [
       {
-        name: 'Import',
+        name: 'import',
         icon: '<i class="fa fa-plus"></i>',
         actions: [
           'upload',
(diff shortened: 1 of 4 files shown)


jczaplew:
This looks great! I say go for it.

Issue #206, opened 2015-10-23
dragging shapefile .zip doesn't load, just hangs


mapsam:
Trying out this shapefile, and it just sits loading forever without any console errors.

ZIP: https://drive.google.com/open?id=0B0E-VGwUGzsSaUtodVVEaWxZWVU

jczaplew:
My guess is that shapefile-js/proj4js doesn't recognize that specific projection and chokes. I'll investigate.

jczaplew:
Do you think we can consider this closed? It's not exactly a satisfying solution, but I was unable to figure out what projections shapefile-js supports ( [issue here](https://github.com/calvinmetcalf/shapefile-js/issues/50) ). Ideally we would make it work with any projection, but I'm not sure what the scope of that job entails.

mapsam:
Yah, let's close for now!

PR #210, opened 2015-10-24
Add some error handling to shapefile loading

jczaplew:
A band aid for #206. Apparently shapefile-js doesn't automatically handle *all* projections, just a lot of common ones. I added some error handling so that if the projection isn't recognized an error is thrown.


mapsam:
Ah, makes sense. Error handler is a good save for now.

Issue #205, opened 2015-10-22
feature request: Zoom to layer extent on click


mapsam:
From https://twitter.com/jlivni/status/654347632576368640

alukach:
So what do you see this as being? I can imagine building out right-click menus for each layer, but that doesn't sound very _dropchoppy_.  We can add a little magnifying class on hover next to the Copy / Delete icons: 

![image](https://cloud.githubusercontent.com/assets/897290/10670045/d5cfb130-78a0-11e5-9b53-bf3aab03f375.png), 

But I don't know how sustainable that is.  As mentioned in #130, I see adding the ability to visually style each layer as well.  We wouldn't want that set of icons to get too busy.

Thourhgts?

mapsam:
I'm not entirely sure. I agree that adding it as a specific option would start to make things a bit cluttered. Might be worth exploring a layer options menu (right click or otherwise, similar to #197) sooner than later so we can add a number of functions to it.

I don't think we should pan/zoom just when the user clicks on a layer though, since that could be a LOT of movement, which could be disorienting. Maybe load to the extent of layers created through operations?

alukach:
Yeah, clicking the layer name is for selecting, not for zooming.  To me, that feels natural and expected.  Above I was saying that it would be another function run on a layer (similar to copy/delete), as either an icon that kicks off the zoom or a right-click menu option.  Or did I misunderstand what you were just saying above?

alukach:
Actually, one important thing to distinguish: when we are saying "click on layer", are we talking about the layer representation in the layerlist or the layer representation on the map?  On layerlist, that currently selects; on map view, that seems to bring up additional metadata about the layer.

mapsam:
Good point! I was talking about layerlist, which I agree should be reserved for just selection. Clicking the map layer up for discussion.

jlivni:
Maybe double clicking to zoom to extent?  But given the other related layer ideas perhaps a right click menu is a better long term plan.

mapsam:
Ohhh double clicking.. that's something that hasn't come up yet. Seems like double clicking would be more of an easter-egg/power-user component, rather than a first-time user UI piece. Worth keeping that in mind for other things as well.

This reminds of an idea @bgolder brought up, where we could building hot-keys for the entire system, allowing for quick power-user experience.

[...]

PR #211, opened 2015-10-26
Right click layerlist layer

alukach:
In response to the comments on #205, here's a first-shot at adding right-click menu support for layers on the layerlist.

This entailed a bit of a refactor in the way menus currently worked.  Each menu (left sidebar, operations menu, and right-click context-menu) was separated out into separate files and namespaces (`dc.menus.left`, `dc.menus.geo`, and `dc.menus.layerContextMenu`).  A `'layerlist:added'` signal was created to return newly added layers. 

Currently, the right-click context menus only work on a single layer.

right-click-layer

There's still work to be done:

- [x] Add support for right-click operations on multiple selected layers.  
- [x] Limit context-menu options based on layers selected (eg 'rename' would only apply to single layer selections but 'zoom to extent' could take many layers). 
- [ ] Write tests
- [x] Remove redundant operations from left sidebar (likely everything under `'Geo Actions'`). 
- [x] Remove icons from layerlist layer item (the 'duplicate' operation should be moved to the right-click menu, the 'remove layer' operation should be removed). _Should this be done?_
- [ ] Add 'right-click' tip to "Welcome to dropchop!" text

I'm opening this up to get feedback.  Is this the right direction for DropChop?


src/js/menus.layer-context-menu.dropchop.js
```diff
@@ -0,0 +1,89 @@
...
+  };
+
+  dc.menus.layerContextMenu._getMenuOperations = function() {
+    return {
+      /* jshint ignore:start */
+      'extent': dc.ops.file['extent'],
+      'expand': dc.ops.file['expand'],
+      'combine': dc.ops.file['combine'],
+      'rename': dc.ops.file['rename'],
+      'remove': dc.ops.file['remove'],
```
alukach: This was written before I saw how #208 rewrote how functions were loaded with the Left Sidebar Menu.  This could probably be refactored to more closely match that style for the sake of simplicity.  One win that this style has is that it could allow operations to be pulled from different files.

alukach:
Supporting multiple layers turned out to be easier than I expected.

right-click-layer

mapsam:
Wowowow this is great! I'm very much into the idea of splitting menus into their own file, since they are growing larger. This makes a lot more sense (i.e. feels more *droppychoppy*) as well - the left-hand "geo actions" felt a bit forced, and very unintuitive since it all related to the layers you had selected anyways. :+1: for pushing this :smiley:

alukach:
@mapsam great, glad you're into it.  Feel free to add some styling changes/suggestions to the menu.

Got validation on the context-menu operations completed:

right-click-layer

alukach:
Update on this:

* Removed icons on side of layers, add an ellipses to indicate "more options". Left-clicking on ellipses button triggers right-click event (for touch-screen users)
* Added headers to the context-menu to allow the operations to be organized. Not sure about this feature, thoughts?
* **problem:** Currently, it's possible for a context menu to go offscreen.  This can occur when right-clicking on layers near the bottom of the layerlist. Not sure what to do about this, adding a scrollbar (and autosizing the context menu) seems difficult.

right-click-menu

mapsam:
This is looking really great.

I think the overflow issue can be solved by doing a little bit of math - checking the height of the menu and the window and ensuring it never generates beyond it. I'd be happy to work on this part sometime this week!

alukach:
@mapsam feel free to take a crack on keeping the menu on the screen. I admit that I hadn't really thought about just comparing the `div` size and placement to the window size and to ensure that it stays on screen, that sounds like a solid solution.

mapsam:
I'll take a look - I'm in my last week out in Richmond for Code for America so might not get to it this week - feel free to try and merge in and I can take a crack in a separate branch!

alukach:
right-click-menu2

Some style changes. Just to add some tests now...

mapsam:
Added the portion to calculate the menu location based on the window size. Seems to be working nicely. I'm a big fan of this menu system, it's much improved @alukach!

dropchop-contextmenuoverflow

mapsam:
@alukach whenever you think this is ready to roll, feel free to merge!

Issue #201, opened 2015-10-20
TopoJSON


jczaplew:
Add the ability to drag and drop TopoJSON, import it from a URL, and export to TopoJSON. This will obviously require the addition of one dependency (TopoJSON), but definitely increases interoperability.

UI-wise I don't think anything should change - it should be very easy to figure out if a file is Geo or TopoJSON, i.e. something like

````
if (data.type === "FeatureCollection") {
  // GeoJSON
} else if (data.type === "Topology") {
  // Topojson, convert to GeoJSON
} else {
  // Hmmm...
}

````

mapsam:
Great idea. Adding a dependency for this is totally acceptable!

jczaplew:
In order to export as TopoJSON, a UI change of some sort will be required. I see two options:

+ Add another button
+ Collapse `Save as GeoJSON` and `Save as Shapefile` down into a single button/menu/modal and add TopoJSON export there.

I'm partial to the latter option. Perhaps a general `Save` button with a submenu that pops out with the three options?

mapsam:
Agreed on the submenu! Seems like a super useful UI component to incorporate as we add new options.

jczaplew:
Export as TopoJSON can be added once #208 is merged

mapsam:
Merged. Ready to roll.

jczaplew:
Turns out saving as topojson isn't as easy as I expected https://github.com/mbostock/topojson/issues/208

mapsam:
Bummer :frowning: - what about at least getting TopoJSON to import? [leaflet-omnivore](https://github.com/mapbox/leaflet-omnivore) would be a solid addition IMO.

jczaplew:
See #212 ;-)

mapsam:
:clap:

PR #212, opened 2015-10-26
TopoJSON

jczaplew:
Added the ability to add and export TopoJSON as outlined in #201. The main changes are as follows:

**Front**
+ Added a button to the `save` menu to save as TopoJSON
+ Added ability to add a TopoJSON file by dragging and dropping, from a gist, or from a URL

**Back**
+ Added a new dependency - `topojson`
+ Added three new dev dependencies - `browserify`, `vinyl-source-stream`, and `vinyl-buffer`
+ In order to use the server API on the client side for saving to TopoJSON, a new script was added `lib/topojson_setup.js` that creates a global object for both the client and server API. However, for now only the `topology` function from the server API was added so that unnecessary methods are not included. It is then run through `browserify` and included in `vendor.js`.

It'd be nice to add tests as well, but it seems like there are not currently tests for this part of DC (though perhaps there should be?). Thoughts?

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -38,6 +38,7 @@ var vendorJS = [
   './node_modules/browser-filesaver/FileSaver.js',
   './node_modules/shp-write/shpwrite.js',
   './node_modules/shpjs/dist/shp.min.js',
+  './node_modules/topojson/topojson.min.js',
   './src/lib/shp-2-geojson.js',
   './node_modules/osmtogeojson/osmtogeojson.js',
   './node_modules/turf/turf.js',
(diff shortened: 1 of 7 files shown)


jczaplew:
It seems like the tests are failing because TopoJSON uses Float64Arrays and phantomjs does not support them because it uses a super old version of webkit - https://github.com/ariya/phantomjs/issues/10607

jczaplew:
I'm not sure why this test failed -- https://travis-ci.org/cugos/dropchop/builds/87390329

I just tried it with node 0.12.7 on my local machine and it worked fine. Has anyone else seen that error?

jczaplew:
There we go! Now that things are passing, any objections to merging into master?

mapsam:
:+1:

PR #217, opened 2015-10-27
Add custom base layer

### LINKED ISSUE #74: Add your own base layer
Now that @lyzidiamond added the feature to switch base layers, it would be awesome to implement a custom layer. This would essentially allow the user to input their own Mapbox tile ID and key?

@lyzidiamond do you know if this would require a key or would the ID be sufficient?

hamhands:
### What changed
* added custom base map button
* takes map id and layer name

### Issues
* Closes #74 

### What's needed 
* way to select the layer once it's added

### Screencap
custom_base

### DIFF
--- a/src/js/ops.file.dropchop.js
+++ b/src/js/ops.file.dropchop.js
@@ -196,8 +196,6 @@ var dropchop = (function(dc) {
       }
     },
 
-    'break1': { type: 'break' },
-
     'save-geojson': {
       minFeatures: 1,
       description: 'Save as GeoJSON',
--- a/src/js/ops.file.dropchop.js
+++ b/src/js/ops.file.dropchop.js
@@ -157,6 +157,47 @@ var dropchop = (function(dc) {
       }
     },
 
+    'load-custom-base': {
+      description: 'Add custom base map',
+      icon: '<i class="fa fa-map-o"></i>',
+      _temp: '',
+      parameters: [
+        {
+          name: 'Map ID',
+          description :'Map ID',
+          type: 'text',
+          default: 'hamhands.npbf8c5p',
+        },
+        {
+          name: 'Custom Base Layer Name',
+          description :'Name of the custom baselayer added.',
+          type: 'text',
+          default: 'METAL MAP!!!'
+        }
+      ],
+      execute: function() {
+        $(dc).trigger('form:file', ['load-custom-base']);
+      },
+      get: function(event, name, parameters) {
+        dc.ops.file['load-custom-base']._temp = parameters[0];
+        dc.ops.file['load-custom-base']._temp_1 = parameters[1];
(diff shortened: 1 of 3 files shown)


powersa:
Make this the default!

alukach:
:suspect: :rage1: :rage2: :rage3: :rage4: :gobeserk:

Love it, great work.

src/js/ops.file.dropchop.js
```diff
@@ -157,6 +157,47 @@ var dropchop = (function(dc) {
...
+          var newLayer = dc.map.baseLayers.addBaseLayer(L.mapbox.tileLayer(dc.ops.file['load-custom-base']._temp),
+            dc.ops.file['load-custom-base']._temp_1);
+          L.mapbox.tileLayer(newLayer).addTo(dc.map.m);
+        } else {
+          dc.notify('error', xhr.status + ': could not retrieve base map. Please check your Map ID');
+        }
+      }
+    },
+
+    'break1': { type: 'break' },
```
alukach: Not sure if this is needed anymore, as we manually construct the menu now.

jczaplew:
It would be nice to be able to add custom base maps from sources other than Mapbox as well. Perhaps we could check for the presence of `http` in the provided URL/map ID and display accordingly? 

Also, since these things are URL-based we could definitely store this state in the URI like we do for gists and layers from URLs

mapsam:
Also, is codeship broken here @alukach? I see all of the open PRs with nothing but nothin on the codeship side.

src/js/ops.file.dropchop.js
mapsam: Are you talking specifically about the `break` piece in there?

alukach: Yup, just that part.

mapsam: Doh, sorry didn't realize this was a line comment.

alukach:
@mapsam not sure what's up with codeship here... seems like it didn't even try to build. If I had to guess, I'd say that it was because this is a PR from @hamhands fork (but that's a total guess).

mapsam:
@hamhands have you worked through loading in non-Mapbox basemaps at all? I'd like to see that!

I think most basemaps that require a token can have the token included in the URL so just having a single field for URL should suffice.

hamhands:
@mapsam haven't yet but I can get going on that! Shouldn't be too big of a change.

mapsam:
Awesome - I'll take a look at selecting the new layer as well. Keep you posted!

mapsam:
Hey @hamhands want to switch out the `xhr` for `$.ajax()` at some point as well? Going to deprecate our own ajax function for the strong jQuery version after some talk in #191

mapsam:
I'm getting a `Uncaught TypeError: Cannot read property '0' of undefined` when the layer adds - it doesn't seem to break anything, but any idea where it's coming from?

hamhands:
Hmm, not sure. Maybe there's an unused variable or incorrect syntax in the execute section?

PR #221, opened 2015-10-27
Let's try TopoJSON again

jczaplew:
Oye...not sure where everything exactly went wrong last time, but it definitely had something to do with the polyfill I added for Phantom. I'm going to leave it out for now.

Can someone pull this down and give it a spin to confirm that it works as expected?

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -9,7 +9,9 @@ var jshint = require('gulp-jshint');
 var ghPages = require('gulp-gh-pages');
 var Server = require('karma').Server;
 var file = require('gulp-file');
-
+var browserify = require('browserify');
+var source = require('vinyl-source-stream');
+var buffer = require('vinyl-buffer');
 
 gulp.task('sass', function(){
   return gulp.src('./src/scss/dropchop.scss')
@@ -41,6 +43,7 @@ var vendorJS = [
   './src/lib/shp-2-geojson.js',
   './node_modules/osmtogeojson/osmtogeojson.js',
   './node_modules/turf/turf.js',
+  './dist/static/js/topojson_package.js',
   './lib/mapbox.js/mapbox.js'
 ];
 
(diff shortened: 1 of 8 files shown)


mapsam:
Pulled down, saved states as topojson and loaded it back in just fine. No weird polygons. :+1:

jczaplew:
Alright, shall we try this again?

mapsam:
Yes, but first figure out these test failures. Looks like phantom is still busting at `ReferenceError: Can't find variable: Float64Array`

jczaplew:
That error has to do with phantom using an old version of webkit (https://github.com/cugos/dropchop/pull/212#issuecomment-151498301), and TopoJSON relying on typed arrays which are a slightly newer feature. I used a polyfill before but it broke the encoding, so I'll have to play around with that a little more.

mapsam:
Ah, you've told me that already. Just needed to hear it twice :grin: 

Would it be worth adding a `Float64Arrays` polyfill into `helper.js` just for the tests? Kind of like [`Function.prototype.bind()`](https://github.com/cugos/dropchop/blob/master/test/helpers.js#L1-L23) already?

mapsam:
Might be able to use this file? https://github.com/inexorabletash/polyfill/blob/master/typedarray.js

jczaplew:
Good call on typedarray.js! It was just a matter of me figuring out how phantom and karma fit together, and what exactly is going on. Way less janky now.

jczaplew:
I think we are good to merge now. Want to just double check the last commit to make sure I didn't break anything?

mapsam:
Huzzah :raised_hands: 

I like how we can use the polyfills straight from `node_modules`. Looks great to me!

jczaplew:
Thanks Sam! Really sorry that I screwed things up for a bit there. I'll be more thorough about testing going forward.

mapsam:
Slipped by me too - and what fun would it be without breaking it a few times? :wink:

PR #224, opened 2015-10-30
Sort available turf functions when layer is selected. Closes #207.

### LINKED ISSUE #207: Push Available Functions to Top of List
Push available functions to top of function list, maintain alphabetical order within available functions.

jczaplew:
So jQuery. Much wow.

Per #207, sort the available geometry functions by availability and alphabetically when a layer is selected. Added a function `_sortGeo` that does the sorting by selecting the buttons and sorting them all alphabetically, then reattaching the active functions followed by the inactive ones. 

If we were to ever add underscore as dependency this could be cleaned up by using a chained `_.sortBy`. Also, there is probably a fancy sorting function you could write to do this all at once, but alas, after playing with a bunch of convoluted methods this worked best and I think is easy to read.

### DIFF
--- a/src/js/ops.dropchop.js
+++ b/src/js/ops.dropchop.js
@@ -142,7 +142,30 @@ var dropchop = (function(dc) {
       throw err;
     }
   }
+
   /* jshint ignore:end */
+
+  // Sort turf operations by availablility and alphabetically
+  function _sortGeo() {
+
+    // Get all of the turf buttons and sort them alphabetically
+    var alpha = $('.operation-geo').sort(function(a, b) {
+      return ($(b).data('operation') < $(a).data('operation')) ? 1 : -1;
+    });
+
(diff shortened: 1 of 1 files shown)


jczaplew:
Hmmm...odd that one of the builds failed. Things passed for me locally, so I'm not sure where to start debugging this one.

jczaplew:
I reran the Travis CI pr and things passed, even though the same error was thrown as the first time. All the tests passed both times, so I'm not sure what the deal is there.

As for the codeship tests, I'm unable to see what went wrong there.

mapsam:
Nice work! Just reran code ship and it worked fine. It's been erroring out randomly more often. Maybe I'll take a look at why sometime.

jczaplew:
With tests passing, I'm going to go ahead and merge.

PR #229, opened 2015-11-07
Hamhands custom basemap

mapsam:
Getting some things working for @hamhands here!

- [ ] any URL, not just mapbox
- [ ] select layer on creation


mapsam:
Stale PR, closing for now @hamhands

Issue #190, opened 2015-10-14
create CONTRIBUTING.md


mapsam:
@lyzidiamond [once said in a PR](https://github.com/cugos/dropchop/pull/70#issuecomment-100392954) that we should make a `CONTRIBUTING.md` file for this. I think it is time!

npeihl:
:+1: 

I'd suggest using https://github.com/jden/CONTRIBUTING.md as starting point.

npeihl:
Ooh, I like this one, too.

https://github.com/CoralineAda/contributor_covenant.

mapsam:
Thanks for the templates, @npeihl - I really like this [code of conduct](http://contributor-covenant.org/version/1/2/0/code_of_conduct.md) piece from the contributor covenant.

mapsam:
@lyzidiamond I'd like to see some examples of what contributing looks like (specifically non-code related contributions). For example, "drawings of new navigation buttons" or "need help creating example data" or "think about color scheme".

Similarly, defining **what is a contribution?** and making it clear that writing code is not the only path towards a successful contribution. Explaining what a meaningful contribution looks like is important! Things like "asking questions" or "requesting features" are equally powerful as creating those features.

THIS IS WHAT DEMOCRACY LOOKS LIKE

![](http://41.media.tumblr.com/782d478fc64bda292a677433fea133af/tumblr_inline_nmlx4zH0iI1tnn1p2_500.jpg)

mapsam:
Some feedback from @hamhands - he wants to contribute but the issue list is a bit overwhelming. Might be worth maintaining a list of comfortable, small tasks in the file so people can start there?

mapsam:
Also, when we ask for code-related contributions how do we expect people to merge into the proper working branch? Should we always assume contributions come into the `master` branch or do we always state where the current development branch is (i.e. `0.1.1` currently)?

mapsam:
Sorry, big brain dump happening right now:

Talking to someone after the presentation at NACIS, they mentioned they were at the CUGOS Spring Fling last spring and enjoyed being a part of the project during the day, but weren't sure how to continue contributing after the fact. I'm not sure if there is a tangible way to answer this issue, but it reminds me a lot of what @lyzidiamond has been talking about with "post user experience" - how do we establish a positive "post contributor experience"?

npeihl:
This looks like a possible way to encourage new contributors. http://yourfirstpr.github.io/

mapsam:
[...]

PR #232, opened 2015-11-09
Put documentation into project files

alukach:
As mentioned in #190, I think we should put as much documentation into the actual repo as possible.  This allows for documentation to be added as part of Pull Requests and pins to the documentation to the checked-out version of code.

## README.md

As features grow, I think it's going to be important to add descriptions of each feature so that users can be made aware of all of Dropchop's potential.  Placing these in the README.md file would probably provide the most visibility of these docs (eventually they should probably be rendered in the application itself). I put in the following sections into the README file:

* Adding Data
  * Upload
  * URL
  * Open Street Map
  * Gist
  * ArcGIS Server Feature Service
  * User Location
* Exporting Data
  * GeoJSON
  * Shapefile

If approved, I'll make an Issue for each of those sections and nudge the developer who implemented that feature.

## CONTRIBUTING.md

I've moved all of the documentation about Dropchop's inner workings (signals, adding turf) into the CONTRIBUTING.md file.  I've also added some notes about Collaborating, Prereqs, Documentation, They feel a bit forced in there and could possibly be moved into a separate file (not sure what that would be called).  Additionally, I feel like describe what is passed with each signal and what each signal is used (that would be another Issue).

## CHANGELOG.md

It seems like the CHANGELOG.md doc is a bit out of date, anyone want to go in and update that from the merged PRs?

## INSTALLATION.md

Moved the installation docs into INSTALLATION.md.

If approved, I'd suggest that we delete the duplicate documentation from the wiki.


mapsam:
Excellent. I think this is great!

alukach:
Great, going to go ahead and merge.

alukach:
Fun fact, adding the `CONTRIBUTING.md` adds this link to all PRs:

screen shot 2015-11-08 at 10 31 47 pm

mapsam:
Ha, awesome.

Issue #239, opened 2015-11-09
Need documentation for adding User Location as a layer


alukach:
We need documentation in README.md for how a user can add their location as a data layer in Dropchop.

mapsam:
Might want to grab @jonahadkins for this one!

jonahadkins:
:+1: i'll do this today sometime

Issue #238, opened 2015-11-09
Need documentation for adding data via Feature Service


alukach:
We need documentation in README.md for how a user can add data via ArcGIS Server Feature Service to Dropchop.

alukach:
@npeihl you added this feature, care to take a swing at adding some documentation?

npeihl:
I'm on it. :notebook_with_decorative_cover:

alukach:
@jonahadkins threw down a basic one-sentence write-up and gif.  If you're up to it, I think there's still room to describe the different fields and possible gis server requirements of the feature.

Issue #237, opened 2015-11-09
Need documentation for adding data via Gist


alukach:
We need documentation in README.md for how a user can add data via Gist to Dropchop.

alukach:
Closed via #244 and #243

Issue #236, opened 2015-11-09
Need documentation for adding data via OSM


alukach:
We need documentation in README.md for how a user can add files via OpenStreetMap to Dropchop.

mapsam:
This one will require a note saying that it isn't feature-complete, yet.

alukach:
Closed via #244 and #243

PR #244, opened 2015-11-10
Patch 1

jonahadkins:
hi! :wave: 
are ya'll wanting gif-only explanations in the readme or gif + text? 
this pr is for issue #239


alukach:
GIFs are definitely a +1. Are there any caveats that a user should be aware of?  This feature is pretty clear in concept so a lot of text probably isn't too crucial, maybe just that it "Adds the users location as a Point feature layer.  The user must grant Dropchop permission to use its location to generate the layer" or something like that?

jonahadkins:
Can Do - i can run some gif's for each function on toolbar if that will help :+1:

npeihl:
I wonder if it would be useful to include a link to the [W3 Geolocation API Specification](http://www.w3.org/TR/geolocation-API/), or is that unnecessary?

jonahadkins:
this pr now includes descriptions and gifs for #234 #235 #236 #237 #238 #239  

should be a good start :+1:

alukach:
Woah! Above and beyond @jonahadkins! Definitely a good start!

@npeihl Not sure how technical we should get. I see this as being a starting location for a user's manual (rather than a developer's manual).  Eventually, we'll probably want to serve this data as a view in the app itself.  Maybe in the future we should think about adding developer's notes to each feature as well?

PR #247, opened 2015-11-17
Update INSTALLATION.md

thedeb:
Updated INSTALLATION.md with some notes tailored for Windows.

### DIFF
--- a/INSTALLATION.md
+++ b/INSTALLATION.md
@@ -1,6 +1,6 @@
 # Installation
 
-These install instructions have only been tested on Mac, updates and comments for Windows are in the works. Please let us know if you have any issues installing on other platforms [here](https://github.com/cugos/dropchop/issues/94).
+These install instructions have only been tested on Mac, updates and comments for Windows are in the works. If you have plroblems installing, please let us know [here](https://github.com/cugos/dropchop/issues/94).
 
 1. **Python** - following steps require [Python](https://python.org/). Current Macintosh systems include Python, Windows users may need to install it. Windows testing was with Python 2.7.5, 32-bit.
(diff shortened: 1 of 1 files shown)


npeihl:
@thedeb It's a Monday, so forgive me if this is a stupid question. Why is Python needed for Dropchop?

thedeb:
Hi Nick,
I thought we saw a dependency when installing npm on Windows, I think it was. @mccombsp-kingco, is that right?
Thank you for checking my work.

mapsam:
Looks like there are some [other python-related install issues on Windows](http://stackoverflow.com/questions/15126050/running-python-on-windows-for-node-js-dependencies). This is a really weird barrier, and really has nothing to do with dropchop at all.

I'm honestly kind of surprised python isn't installed on Windows machines natively?

mccombsp-kingco:
Microsoft doesn't expect all users to be developers. Esri users will have
it installed, most others won't.

npeihl:
@mccombsp-kingco I guess that's why I didn't have any issues. I was testing it on my work machine (Esri software installed). Thanks.

mapsam:
Good point!

Seems like most of the issues out there with Windows and `npm` deal with [`node-gyp`](https://www.npmjs.com/package/node-gyp) and the environment variable not existing. Not sure if there's anything we can do other than make a comment in the install docs like @thedeb has already done.

npeihl:
I'm satisfied with @thedeb's and @mccombsp-kingco's  work and suggest we merge. :100:

mapsam:
Agreed! :100: 

Looks like we are getting a timeout due to the size of `js_vendor` in `gulp-uglify`. Not entirely sure how to prevent this.

mapsam:
Confirmed the error exists for all tests, in #248

Issue #249, opened 2015-12-07
Support 'ctrl' key for multiple layer select in Ubuntu


powersa:
Use control c to select multiple layers that aren't next to each other.

alukach:
Is this not what currently happens? By holding control, I believe you can
select multiple layers within the layer list regardless if they're
contiguous or not. Adding 'shift' will select all layers between current
selection and clicked layer.

mapsam:
Indeed. Some screenshots of what works for me:

Using `command` key on Mac
![dropchop-selection-command](https://cloud.githubusercontent.com/assets/1943001/11638056/a7032762-9cd9-11e5-866f-418fea55c006.gif)

Using control key on mac opens menu
![dropchop-selection-control](https://cloud.githubusercontent.com/assets/1943001/11638057/a703d86a-9cd9-11e5-8212-2ba9746c0b88.gif)

Using shift selects anything in between
![dropchop-selection-shift](https://cloud.githubusercontent.com/assets/1943001/11638058/a7049fca-9cd9-11e5-9f98-c90f475ee600.gif)

mapsam:
@powersa are you using a PC? I wonder if special selections are different between Mac and PC

powersa:
Woo, this might be something really specific to my setup. I'm running ubuntu 14.04 on a Lenovo ThinkPad T440s. The ctrl key doesn't work for me. But my computer does have a windows key that lets me select non-contiguous layers. 

Anybody else on ubuntu able to test?

mapsam:
Hm! Interesting. Let's test this on Wednesday at the sidecar.

powersa:
sounds great, thanks @mapsam

alukach:
@powersa I use Ubuntu at work but don't experience this issue. Which browser are you using?  It's strange that <kbd>ctrl</kbd> key doesn't do its magic on your system.

alukach:
Ah, disregard my above comment, the <kbd>ctrl</kbd> doesn't work on my Ubuntu machine! Looks like we've got a legit bug on our hands, thanks @powersa. (For the record, using Chrome Version 46.0.2490.86 (64-bit) on Ubuntu 14.04).

powersa:
oh my oh my. such relief. thanks for following up y'all.

mapsam:
Will take a look at the CUGOS sidecar tomorrow night :+1:

alukach:
[...]

PR #250, opened 2015-12-08
Allow 'ctrl' key to be used for select-many

alukach:
I believe that this should do the trick for the issue raised by @powersa in #249.

Note: I haven't actually tested this, just read through the code and made the PR via my browser.


mapsam:
:clap: 

Some information about browsers and OS differences: http://stackoverflow.com/questions/6245868/which-key-does-the-e-metakey-refer-to-in-javascript-mouseevent

PR #252, opened 2016-01-17
use overpass-wizard to generate Overpass API queries

tyrasd:
This enables a human friendly syntax to query for almost arbitrary overpass queries. A few examples:
* amenity=bar in "New York City"
* building=*
* type:relation and route=bicycle

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -49,7 +49,9 @@ var vendorJS = [
   './node_modules/shp-write/shpwrite.js',
   './node_modules/shpjs/dist/shp.min.js',
   './node_modules/turf/turf.js',
-  './src/lib/shp-2-geojson.js'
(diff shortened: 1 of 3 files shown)


mapsam:
Wow! Terrific work @tyrasd. This is a HUGE upgrade to dropchop. Everything looks good to me here, nice work getting this all added as a node module - exciting to see the wizard as its own project.

The only thing I can see here that will need some work (outside of this PR) is us standardizing the `xhr` requests. I think we tried to get everything moved to `$.ajax()` since we are already using jQuery in #191 but I don't know if we ever removed the `xhr` utility. I'll open another issue to clean this out.

Onward!

mapsam:
Confirmed with a sweet gif:

dropchop-overpasswizard

PR #256, opened 2016-07-20
toggleLayers update

SwampGuzzler:
The 'Toggle all layers' checkbox actually runs counter to how it should. On initial load, if we have layers loaded from the hash, the first 'uncheck' of that checkbox actually does nothing. The jquery checks that were being run in the 'dc.layerlist.toggleAll' worked after that, but since the initial 'loaded' property on the 'Toggle all layers' input was now gone, the toggleAll checkbox would actually be checked when the map layers were hidden, and unchecked when they were displayed.
For this particular use case, document.getElementById is actually a safer bet, hence the new 'allCheckboxes' id.

I also updated the topojson path because it failed on our 'js_vendor' gulp task.

### DIFF
--- a/lib/topojson_setup.js
+++ b/lib/topojson_setup.js
@@ -1,4 +1,4 @@
 window.topojson = {
   topology: require('../node_modules/topojson/lib/topojson/topology'),
-  client: require('../node_modules/topojson/topojson.js')
+  client: require('../node_modules/topojson/build/topojson.js')
 }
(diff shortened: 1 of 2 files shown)


SwampGuzzler:
Also, the travis-ci check fails on 'gulp build:prod' due to a 'Maximum call stack size exceeded' error, rather than any karma-run unit test failures.

mapsam:
Heh, funny bug - thanks for working on this @SwampGuzzler! I'll take a look this weekend and will look into solving the travis timeouts as well.

mapsam:
Fixed in https://github.com/cugos/dropchop/pull/257

PR #257, opened 2016-07-30
fix travis tests

mapsam:
Working on fixing the tests in travis - currently running into `Maximum call stack size exceeded` errors during our vendor script, which takes a long time.


mapsam:
Huzzah! Removing our custom mapbox.js bundle and just going straight to the CDN has improved travis speeds immensely. @wboykinm @SwampGuzzler 

Unfortunately, the topojson client is a bit different these days and an attempt at babelifying it doesn't result in code that works in browser, and still tries to `fs.readFileSync`. I've copied over our current `topojson_package.js` in the `lib/` so we can continue to use it until we can dig further. cc @jczaplew 

TODO:

- [ ] fix codeship

mapsam:
Ah, just a path change needed for topojson - thanks @SwampGuzzler!

mapsam:
Woo! This is good to go.

Issue #261, opened 2016-07-31
Modularize turf dependencies


wboykinm:
Instead of requiring turf whole-cloth, just install the functions that are configured for dropchop. Currently:

```
"turf-along": "^3.0.12",
"turf-bezier": "^3.0.12",
"turf-buffer": "^3.0.12",
"turf-center": "^3.0.12",
"turf-centroid": "^3.0.12",
"turf-destination": "^3.0.12",
"turf-envelope": "^3.0.12",
"turf-explode": "^3.0.12",
"turf-midpoint": "^3.0.12",
"turf-simplify": "^3.0.12",
"turf-tin": "^3.0.12",
"turf-union": "^3.0.12",
"turf-within": "^3.0.12"
```

wboykinm:
@nickpeihl I'm (understandably) getting `turf is not defined` [here](https://github.com/cugos/dropchop/blob/master/src/js/ops.geo.dropchop.js#L109) when I run dropchop locally with only the modules installed (instead of the whole package). 

The parent object `turf` seems to be required, but (many apologies) I can't find where it's defined in the codebase to replace it. Adding it back [to the vendorJS array](https://github.com/cugos/dropchop/blob/master/gulpfile.js#L45) points to a nonexistent node module, so that doesn't work. Other things that don't work include:

- `var result = turf-buffer.apply(null, params);`
- `var result = buffer.apply(null, params);`
- `var result = turfBuffer.apply(null, params);`

mapsam:
@wboykinm I believe that, since we are still not fully using browserify to build dropchop, you'll have to do a similar thing that @jczaplew did with topojson that:

1.  [includes a source file in `lib`](https://github.com/cugos/dropchop/blob/master/lib/topojson_setup.js) that requires the necessary modules
1. and a [browserify command in gulp](https://github.com/cugos/dropchop/blob/master/gulpfile.js#L66-L73) that builds the script for dropchop before ...
1. [loading into the js_vendor](https://github.com/cugos/dropchop/blob/master/gulpfile.js#L46) command

wboykinm:
All set here.

PR #264, opened 2016-08-03
Modularize

wboykinm:
Per #261, replace turf-the-monolith with individual modules as they are configured.


wboykinm:
@mapsam This looks GTG.

mapsam:
Hey! Nice work @wboykinm 😄 

Updated docs ❤️ 

For the sake of semver, I think staying within `0.2.x` is fine here, since we are only changing dependencies, and functionality won't change/update at all for the end user. `0.2.3` should be 👍

mapsam:
Also sent you an invite to the CUGOS org and dropchop team so you don't have to keep forking!

wboykinm:
I have the worst time speaking semver :)

wboykinm:
@mapsam Okay - bumped back the version. Thanks for the invite! I'll make future changes from upstream.

mapsam:
Beautiful. Merging 👍

mapsam:
Confirmed we are live!

screen shot 2016-08-03 at 10 06 33 am

Issue #265, opened 2016-08-03
Add turf.sample


wboykinm:

wboykinm:
GTG. 

![sample](https://www.dropbox.com/s/ufnoymjra9lizae/sample.gif?dl=1)

PR #266, opened 2016-08-04
Sample

wboykinm:
Add `turf-sample` to the configured functions, per #265

### DIFF
--- a/lib/turf_setup.js
+++ b/lib/turf_setup.js
@@ -8,6 +8,7 @@ window.turf = {
   envelope: require('../node_modules/turf-envelope/index.js'),
   explode: require('../node_modules/turf-explode/index.js'),
   midpoint: require('../node_modules/turf-midpoint/index.js'),
+  sample: require('../node_modules/turf-sample/index.js'),
   simplify: require('../node_modules/turf-simplify/index.js'),
   tin: require('../node_modules/turf-tin/index.js'),
   union: require('../node_modules/turf-union/index.js'),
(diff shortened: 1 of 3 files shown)


wboykinm:
@mapsam I was just digging through tests and noticed that none in the spec seem to operate on the turf functions themselves. Is there a handler or helper for input/output geojson? Even something as simple as getting a response for `expect(outGeoJson.features.length).to.equal(whatever);`?

nickpeihl:
@wboykinm I think best practices says that the owners of the Turf library should be responsible for testing the Turf functions and we should give them the benefit of the doubt on the validity of those functions.

wboykinm:
@nickpeihl Coolness - what then is the policy around here about merging one's own PRs?

nickpeihl:
The only policy we have is http://cugos.org/code-of-conduct/ which doesn't address this specifically. This isn't a breaking change so I say you're gtg if you are happy with it.

mapsam:
Lovely. @wboykinm it's always nice to have a gif/screenshot of what's being added - if that's something you're able to do!

wboykinm:
sample

wboykinm:
Whoops. Totally forgot the changelog and version. Coffee hasn't sunk in yet.

mapsam:
@wboykinm since we don't have a completely robust release mechanism - I've, more than once, committed the version/changelog updates straight into master. shhhh 😄 (Don't think less of me!)

wboykinm:
@mapsam I was hoping exactly the same. Some devops engineer is scowling at me right now, surely.

PR #267, opened 2016-08-05
Revert "Sample"

wboykinm:
Reverts cugos/dropchop#266 to update changelog/version.


wboykinm:
That was the wrong way to do that 👎

PR #269, opened 2018-01-22
bump jquery to 3.3

wboykinm:
Per [the security alert](https://github.com/cugos/dropchop/network/dependencies#30523787) I just plugged in the jquery upgrade. Note that tests are currently busted behind a phantomjs failure, so I just tried it out locally and it "worked fine"™.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "dropchop",
-  "version": "0.2.4",
+  "version": "0.2.5",
   "description": "GIS in the browser.",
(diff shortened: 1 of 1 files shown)


nickpeihl:
LGTM. 

I attempted to replace PhantomJS with [Puppeteer](https://github.com/karma-runner/karma-chrome-launcher#headless-chromium-with-puppeteer) which I hoped would resolve the CI tests, but I had a lot of troubles with outdated dependencies. Maybe when I have more time.

